PR #2, opened 2024-01-18
Local database for caching SBOM data

jhoward-lm:
# Pull Request Template

## Description

Add database schema creation scripts for review.

## Type of change

- [ ] ~~Bug fix (non-breaking change which fixes an issue)~~
- [X] New feature (non-breaking change which adds functionality)
- [ ] ~~Breaking change (fix or feature that would cause existing functionality to not work as expected)~~
- [ ] ~~This change requires a documentation update~~

## How Has This Been Tested?

- [ ] Locally ran `go run main.go` and verified no errors
- [ ] Locally verified generated tables and indexes using `sqlite3` 

## Checklist

- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~
- [ ] ~~New and existing unit tests pass locally with my changes~~
- [ ] ~~Any dependent changes have been merged and published in downstream modules~~
- [ ] ~~I have checked my code and corrected any misspellings~~


jhoward-lm:
> A couple questions to understand better.
> 
> * How was the schema generated? from proto-bom?

It was basically generated from [sbom.proto](https://github.com/bom-squad/protobom/blob/main/api/sbom.proto) by hand, but with some regex find and replace. This is the starting point/proof-of-concept; one possible next progression step might be to represent these tables as `gorm` models. The ideal would be to find a way to dynamically generate the `gorm` SQLite models by compiling the protos with `protoc-gen-gorm` or something.

> * I'm guessing there is a need for a top level "bom" table, is there one?
>   
>   * this could include stuff like bom "aliases", where to find the original bom on disk ...

The top level proto message is `Document`; still kind of working out how to represent this relationship. I was thinking of creating a SQL view that joins the `metadata` and `node_list` tables. But yeah, alias and file path would also be valuable data to include. I'll add those

jhoward-lm:
@idunbarh The closest thing I've found so far to being able to automate the database creation would be [protoc-gen-gorm](https://github.com/infobloxopen/protoc-gen-gorm). There are a couple of potential problems with this:

- would probably require modifications to the original protobom `sbom.proto` file to compile correctly
  - we could maintain our own `sbom.proto` file with options for `protoc-gen-gorm`, but we'd be on the hook to keep it synchronized with protobom's 
  - we could ask protobom to create that additional `.proto` file in their project, but they would be on the hook to maintain it, and might not be willing to

@puerco Do you have any ideas about how to keep a database representation of the protobom models synchronized with the latest protobom release?

howardj99:
I’ve gone back and forth on this one, but it could make more sense to just store the raw bytes/blob data of the protobom buffer in the database and serialize/deserialize as needed

internal/pkg/db/scripts/types.sql
```diff
@@ -0,0 +1,374 @@
...
+
+-- @label Populate values
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (0, 'UNKNOWN');
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (1, 'amends');
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (2, 'ancestor');
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (3, 'buildDependency');
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (4, 'buildTool');
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (5, 'contains');
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (6, 'contained_by');;
+INSERT OR IGNORE INTO edge_type (id, name) VALUES (7, 'copy');
```
puerco, line 324: Strictly speaking, we don't need the enum labels to build the cache. If they get generated automatically it's fine but if we're maintaining the SQL manually it would make sense to drop these.

jhoward-lm:
Instead of using SQL scripts for instantiation, created a fork of the protobom repo and updated the [message definitions](https://github.com/jhoward-lm/protobom/blob/protoc-gen-gorm/api/sbom.proto). Then the tables can be created directly from the compiled protobufs [like this](https://github.com/bomctl/bomctl/pull/2/commits/cb9c309d211ad6b43e32976874c9f7e0126bf18d#diff-5b9beda2c8ccfb609426b8b76a723c56cb66d0cbbca293d7f80838b48af68e0d) instead

jhoward-lm:
@idunbarh Been working with @puerco on it, I'm planning to submit a PR of my fork back into the upstream protobom repo

puerco:
Thanks for the improvements @jhoward-lm, they are very welcome to make our protos more useful 🚀

PR #22, opened 2024-04-18
fix: implementing scorecard suggestions

idunbarh:
Updates to address some scorecard check findings and add badges to the readme

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,12 +60,22 @@ jobs:
 
           echo "next-version=${next_version}" >> "$GITHUB_OUTPUT"
 
-      - name: Push tag for next release
+      - name: Check if tag exists
         if: fromJSON(env.should_publish)
+        id: tag-exists
+        env:
+          NEXT_VERSION: ${{ steps.next-version.outputs.next-version }}
+        run: |
+          [[ -n $(git tag --list "$NEXT_VERSION") ]] && tag_exists="true" || tag_exists="false"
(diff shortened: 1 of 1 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -30,15 +30,14 @@ on:
...
-  pull-requests: read
-  checks: write
+permissions: read-all
 
 jobs:
   golangci:
     name: lint
     runs-on: ubuntu-latest
+    permissions:
+        contents: write # needed to write releases
```
jhoward-lm, line 40: Minor style nitpick

```suggestion

    permissions:
        contents: write # needed to write releases

```

jhoward-lm (approved):

.github/workflows/shellcheck.yml
jhoward-lm, line 1: Just out of curiosity, does replacing scoped read access in the workflows with global read access really improve the scorecard score?

.github/workflows/release.yml
jhoward-lm, line 1: I think this job needs an additional step to check if the tag already exists and save it as an output. Then it can use the new step output to decide whether or not to push a new tag

I’ll write out the job and add it as a suggestion when I get a chance

jhoward-lm (approved):

.github/workflows/release.yml
jhoward-lm: That can be a separate PR if you’re good with it; when this merges, the job should try to create and push a new tag `v0.1.1` if the merge commit has the `fix:` commit type

PR #54, opened 2024-05-21
Adding Custom SBOM Metadata & GitHub Attestation Workflows

idunbarh:
This PR implements two changes to the CI.

1. Updates SBOM generation in `goreleaser` to use a shell script. This shell script calls syft and then merges sbom metadata from `.github/sbom_metadata/metadata.cyclonedx.json` or `.github/sbom_metadata/metadata.spdx.json` to produce the final sbom.
2. Adds the new [GitHub Attestation Store](https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/) to the `goreleaser` workflow. Using this workflow creates SLSA Attestations for the release artifacts and archives. These are also signed using the public sigstore's rekor.

Not super thrilled with how the spdx sboms look, but the final cdx sboms look great!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,6 @@
 !/.vscode/settings.json
 !/.vscode/tasks.json
 !/docs/CHANGELOG.md
-!/internal/pkg/db/testdata/sbom.*.json
 
 # Whitelist patterns found anywhere in project
 !*.go
@@ -38,5 +37,6 @@
 !Makefile
 !README.md
 !SECURITY.md
+!testdata/
 
 !*/


.github/workflows/goreleaser.yml
```diff
@@ -65,9 +66,23 @@ jobs:
...
+        if: startsWith(github.ref, 'refs/tags/')
+        uses: actions/attest-build-provenance@173725a1209d09b31f9d30a3890cf2757ebbff0d # v1.1.2
+        with:
+          subject-path: "dist/*[.zrda][gipep][zpbk]" # match .gz, zip, deb, apk
+
+      - name: Attest Artifacts
+        if: startsWith(github.ref, 'refs/tags/')
+        uses: actions/attest-build-provenance@173725a1209d09b31f9d30a3890cf2757ebbff0d # v1.1.2
+        with:
+          subject-path: "dist/bomctl*/*" # match all binaries created
```
jhoward-lm, line 79:
```suggestion
          subject-path: dist/bomctl*/* # match all binaries created
```

.github/workflows/goreleaser.yml
```diff
@@ -65,9 +66,23 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SNAPSHOT: ${{ github.ref_type != 'tag' && '--snapshot' || '' }}
 
+      - name: Attest Archives
+        if: startsWith(github.ref, 'refs/tags/')
+        uses: actions/attest-build-provenance@173725a1209d09b31f9d30a3890cf2757ebbff0d # v1.1.2
+        with:
+          subject-path: "dist/*[.zrda][gipep][zpbk]" # match .gz, zip, deb, apk
```
jhoward-lm, line 73:
```suggestion
          subject-path: dist/*[.zrda][gipep][zpbk] # match .gz, zip, deb, apk
```

.gitignore
```diff
@@ -12,6 +12,8 @@
 !/.markdownlintignore
 !/.pre-commit-config.yaml
 !/docs/CHANGELOG.md
+!/.github/sbom_metadata/metadata.cyclonedx.json
+!/.github/sbom_metadata/metadata.spdx.json
```
jhoward-lm, line 16: These should go in `.github/.gitignore` as

```
!sbom_metadata/metadata.cyclonedx.json
!sbom_metadata/metadata.spdx.json
```

.github/scripts/sbom-generation.sh
```diff
@@ -0,0 +1,21 @@
...
+}
+
+if [[ $# -ne 3 ]]; then
+  usage
+else
+  syft scan "${1}" --output "${2}-json=syft-${3}"
+
+  # Use jq to merge additional metadata into the sboms
+  jq -s '.[0] * .[1]' "syft-${3}" "../.github/sbom_metadata/metadata.${2}.json" > "${3}"
+fi
```
jhoward-lm, line 21: Since `usage` exits, no need for `else`

```suggestion
if [[ $# -ne 3 ]]; then usage; fi

syft scan "${1}" --output "${2}-json=syft-${3}"

# Use jq to merge additional metadata into the sboms
jq --slurp '.[0] * .[1]' "syft-${3}" "../.github/sbom_metadata/metadata.${2}.json" > "${3}"
```

jhoward-lm (changes requested):

.gitignore
jhoward-lm: This is mostly based on personal experience and opinion, but the original goal was:

- The project level .gitignore uses the allow list template, and only includes patterns anchored to the project root or found anywhere in the project
- The nested .gitignore file(s) are for whitelisting file patterns scoped to that directory and its subdirectories. It keeps everything modular, scoped, and tidy
- A single top level .gitignore can become very messy very quickly, and the patterns tend to become complicated due to the limited expression syntax of file globs
- Since everything is ignored except what is explicitly allowed, it's meant to reduce the chance of developers committing things they didn't intend to. There are of course things like pre-commit for this but not everyone uses it

I use similar patterns for things like Makefiles to keep everything contained to its enclosing scope. It's just the way I like to organize projects to keep my workspace clean (I typically have a lot of untracked files for testing or ideas I want to hold onto for later). If the overall consensus is a preference to use a single .gitignore file, I won't fight it, but I do think we need to do only one way or the other.

jhoward-lm: @eddiezane @ashearin do you have any preferences on this?

ashearin: My preference for this repo would be a single `.gitignore`. Mostly for visibility for users and ease of maintenance: at least in a sense of intuitiveness for contributors. Just my opinion so I could be off base. I do agree that it could become messy though, something to watch. 

I could see a better argument for multiple if this was going the monorepo route, but I don't think thats likely. My feeling aren't strong, and I see advantages for both, so I'm good either way.

jhoward-lm: @idunbarh here's one example

https://github.com/CycloneDX/specification/blob/master/docgen/json/.gitignore

jhoward-lm: Sounds like the single top-level .gitignore gets the votes. @idunbarh do you want to make that change in this PR? I think this is all of them:

- internal/pkg/db/testdata/.gitignore
- .github/.gitignore
- .vscode/.gitignore

.gitignore
```diff
@@ -11,7 +11,18 @@
...
+!/.github/scripts/*.sh
+!/.github/workflows/*.yml
+!/.vscode/extensions.json
+!/.vscode/header.code-snippets
+!/.vscode/launch.json
+!/.vscode/settings.json
+!/.vscode/tasks.json
 !/docs/CHANGELOG.md
+!/internal/pkg/db/testdata/sbom.*.json
```
jhoward-lm, line 25: The `testdata` folder is common in go project subdirectories and automatically ignored by `go build`. So I think this pattern might be cleaner as just `!testdata/` under the "patterns found anywhere in the project" section

It won't let me use suggestions on code that wasn't changed at the bottom of the file

jhoward-lm (approved):

jhoward-lm (approved):

ashearin (approved):

PR #75, opened 2024-06-11
temporary Save command - Save SBOM to File By ID

houdini91:
# Pull Request Template

## Description

Suggestion for a `Save` Command, Considering @jhoward-lm [Comment](https://github.com/bomctl/bomctl/pull/33#issuecomment-2068182831).

Typically, users populate the database using the fetch command. This new command will allow users to save SBOMs to their local filesystem.

## Usage suggestion.
```yaml
Save SBOM file(s) from Storage to Filesystem

Usage:
  bomctl save [flags] SBOM_ID...

Flags:
  -e, --encoding string   the output encoding [spdx: [text, json] cyclonedx: [json] (default "json")
  -f, --format string     output format [spdx, spdx-2.3, cyclonedx, cyclonedx-1.0, cyclonedx-1.1, cyclonedx-1.2, cyclonedx-1.3, cyclonedx-1.4, cyclonedx-1.5]
  -h, --help              help for save

```

## Issues
1) Handling multiple SBOM IDs with a single output file:
Option 1: Output as tar archive
Option 2: Append index to file path
Option 3: Do not support multiple SBOM IDs in the command
Open to other suggestions
 
## Type of change

## How Has This Been Tested?
Waiting for feedback before implementing tests.

### DIFF
--- a/internal/pkg/utils/format/format.go
+++ b/internal/pkg/utils/format/format.go
@@ -1,0 +1,94 @@
+package format
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/protobom/protobom/pkg/formats"
+)
+
+var (
+	DefaultEncoding         = formats.JSON
+	DefaultSPDXJSONVersion  = formats.SPDX23JSON
+	DefaultSPDXTVVersion    = formats.SPDX23TV
+	DefaultCycloneDXVersion = formats.CDX15JSON
+	JSONFormatMap           = map[string]formats.Format{
+		"spdx":     formats.SPDXFORMAT,
+		"spdx-2.2": formats.SPDX22JSON,
+		"spdx-2.3": formats.SPDX23JSON,
+
+		"cyclonedx":     formats.CDXFORMAT,
+		"cyclonedx-1.0": formats.CDX10JSON,
+		"cyclonedx-1.1": formats.CDX11JSON,
+		"cyclonedx-1.2": formats.CDX12JSON,
+		"cyclonedx-1.3": formats.CDX13JSON,
+		"cyclonedx-1.4": formats.CDX14JSON,
+		"cyclonedx-1.5": formats.CDX15JSON,
+	}
+
+	TVFormatMap = map[string]formats.Format{
+		"spdx":     formats.SPDXFORMAT,
+		"spdx-2.2": formats.SPDX22TV,
+		"spdx-2.3": formats.SPDX23TV,
+	}
+
+	XMLFormatMap = map[string]formats.Format{}
+
+	JSONEncoding = formats.JSON
+	TEXTEncoding = formats.TEXT
+	SPDX         = formats.SPDXFORMAT
+	CDX          = formats.CDXFORMAT
+
+	EncodingMap = map[string]string{
+		"json": formats.JSON,
+		"xml":  formats.XML,
+		"text": formats.TEXT,
+	}
+)
+
+type Format struct {
+	formats.Format
+}
+
+// Parse parses the format string into a formats.Format
+func Parse(fs string, encoding string) (formats.Format, error) {
+	if fs == "" {
+		return formats.EmptyFormat, errors.New("no format specified")
+	}
+
+	s := strings.ToLower(fs)
+	var fm map[string]formats.Format
+
+	switch encoding {
+	case formats.JSON:
+		fm = JSONFormatMap
+	case formats.TEXT:
+		fm = TVFormatMap
+	case formats.XML:
+		fm = XMLFormatMap
+	default:
+		return formats.EmptyFormat,
+			fmt.Errorf("unknown encoding: %s", encoding)
+	}
+	f, ok := fm[s]
+	if !ok {
+		return formats.EmptyFormat, fmt.Errorf("unknown format: %s", fs)
+	}
+
+	if f == formats.SPDXFORMAT {
+		if encoding == formats.JSON {
+			return DefaultSPDXJSONVersion, nil
+		}
+
+		if encoding == formats.TEXT {
+			return DefaultSPDXTVVersion, nil
+		}
+	}
+
+	if f == formats.CDXFORMAT {
+		return DefaultCycloneDXVersion, nil
+	}
+
+	return f, nil
+}
--- a/internal/pkg/utils/format/format_test.go
+++ b/internal/pkg/utils/format/format_test.go
@@ -1,0 +1,55 @@
+package format_test
+
+import (
+	"testing"
+
+	"github.com/bomctl/bomctl/internal/pkg/utils/format"
+	"github.com/google/go-cmp/cmp"
+	"github.com/protobom/protobom/pkg/formats"
+)
+
+func Test_Parse(t *testing.T) {
+	tests := []struct {
+		name     string
+		fs       string
+		encoding string
+		want     formats.Format
+		wantErr  bool
+	}{
+		{
+			name:     "Parse spdx-2.2 json format",
+			fs:       "spdx-2.2",
+			encoding: formats.JSON,
+			want:     formats.SPDX22JSON,
+			wantErr:  false,
+		},
+		{
+			name:     "Parse spdx-2.3 json format",
+			fs:       "spdx-2.3",
+			encoding: formats.JSON,
+			want:     formats.SPDX23JSON,
+			wantErr:  false,
+		},
+		{
+			name:     "Parse spdx json format",
+			fs:       "spdx",
+			encoding: formats.JSON,
+			want:     format.DefaultSPDXJSONVersion,
+			wantErr:  false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := format.Parse(tt.fs, tt.encoding)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Parse() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			if !cmp.Equal(got, tt.want) {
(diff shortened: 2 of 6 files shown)


jhoward-lm:
After giving it some thought, I think it makes more sense to me for this sub command to be called `export`, where if no positional parameters are provided it should write to `stdout` for piping to other tools. I still think the `fetch` option should be called `--save` however. Thoughts?

The reason is for parity with a potential `import` command I'm kicking around, which would basically just be `fetch` except designed for local files and `stdin` input.

jhoward-lm:
> If the `fetch` command allows fetching of externalRefs, I think a `save` command that could put "tree structures" as sbom files.

What are your thoughts for the naming convention of the files in that tree structure, or ways to specify multiple output file names and map them to the input URLs

> If the `import` command allows a single SBOM to be piped in, I would expect an `export` to also pipeline a single SBOM.

It might not necessarily have to be a single input or output SBOM, thinking along the lines of how tools like `xargs` handle these situations. Input or output could basically be a whitespace delimited series of multiple super long SBOM content strings

houdini91:
In my opinion, large stdout and stdin strings are useful when using pipes but not very practical when printed directly to the terminal.

I also agree that the save command should be renamed to export. IMHO one if the use case is exporting an SBOM to a third-party system or generating a report. In such cases, outputting a collection/tree of SBOMs files may be useful and more natural to some systems.

I was under the impression that creating a unified SBOM should be the responsibility of the merge command, not the export/save command.

houdini91:
Sorry for the delay, i was yet on another small vacation.

> After giving it some thought, I think it makes more sense to me for this sub command to be called export, where if no positional parameters are provided it should write to stdout for piping to other tools. I still think the fetch option should be called --save however. Thoughts?

Changed `save` to `export` command as suggested.

> It might not necessarily have to be a single input or output SBOM, thinking along the lines of how tools like xargs handle these situations. Piped I nput or output could basically be a whitespace delimited series of multiple super long SBOM content strings

>If the import command allows a single SBOM to be piped in, I would expect an export to also pipeline a single SBOM.

 I Patched the code with both these suggestions, what do you think?
 
 * When uses `output-file` option only one SBOM id is allowed for command.
 * When not using `output-file` option all Sboms are pushed out to stdout.
 
 @jhoward-lm @idunbarh

internal/pkg/utils/format/format.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/protobom/protobom/pkg/formats"
+)
+
+var (
```
jhoward-lm, line 11: The linter doesn't allow global variables. I'd suggest to refactor the maps into functions, and instead of assigning constants like `formats.JSON` to variables, just use the constants inline since they're already bound to a name.

internal/pkg/utils/format/format.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		if encoding == formats.TEXT {
+			return DefaultSPDXTVVersion, nil
+		}
+	}
+
+	if f == formats.CDXFORMAT {
+		return DefaultCycloneDXVersion, nil
+	}
```
jhoward-lm, line 91: How would you feel about something like

```suggestion
	switch f, ok := fm[s]; {
	case !ok:
		return formats.EmptyFormat, fmt.Errorf("unknown format: %s", fs)
	case f == formats.SPDXFORMAT && encoding == formats.JSON:
		return DefaultSPDXJSONVersion, nil
	case f == formats.SPDXFORMAT && encoding == formats.TEXT:
		return DefaultSPDXTVVersion, nil
	case f == formats.CDXFORMAT:
		return DefaultCycloneDXVersion, nil
	}
```

internal/pkg/utils/format/format_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := format.Parse(tt.fs, tt.encoding)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Parse() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+
+			if !cmp.Equal(got, tt.want) {
```
jhoward-lm, line 50: Could this not just be

```suggestion
			if got != tt.want {
```

jhoward-lm (changes requested):
Need to run golangci-lint locally and address findings

cmd/export.go
```diff
@@ -0,0 +1,146 @@
...
+
+	sbomIDs := SBOMIDSliceValue{}
+
+	exportCmd := &cobra.Command{
+		Use:  "export [flags] SBOM_URL...",
+		Args: cobra.MinimumNArgs(1),
+		PreRun: func(_ *cobra.Command, args []string) {
+			for _, arg := range args {
+				sbomIDs = append(sbomIDs, arg)
+			}
```
jhoward-lm, line 48:
```suggestion
			sbomIDs = append(sbomIDs, args...)
```

jhoward-lm (changes requested):

houdini91:
@idunbarh @jhoward-lm I think i fixed all your comments.
Have another look at it.

I ran a local test and it worked as expected.
```
go run main.go fetch https://raw.githubusercontent.com/protobom/protobom/main/test/conformance/testdata/cyclonedx/1.5/json/bom-1.5.json --cache-dir .tmp

go run main.go export urn:uuid:3e671687-395b-41f5-a30f-a58921a69b79 --cache-dir .tmp
```

houdini91:
Current Status of usage.
```
Export SBOM file(s) from Storage

Usage:
  bomctl export [flags] SBOM_URL...

Flags:
  -e, --encoding STRING    the output encoding [spdx: [text, json] cyclonedx: [json] (default json)
  -f, --format STRING      Output Format:
                           	- spdx,
                           	- spdx-2.3,
                           	- cyclonedx,
                           	- cyclonedx-1.0,cyclonedx-1.1,
                           	- cyclonedx-1.2,
                           	- cyclonedx-1.3,
                           	- cyclonedx-1.4,
                           	- cyclonedx-1.5
                           	 (default cyclonedx-1.5)
  -h, --help               help for export
  -o, --output-file FILE   Path to output file

Global Flags:
      --cache-dir string   cache directory [defaults:
                           	Unix:    $HOME/.cache/bomctl
                           	Darwin:  $HOME/Library/Caches/bomctl
                           	Windows: %LocalAppData%\bomctl]
      --config string      config file [defaults:
                           	Unix:    $HOME/.config/bomctl/bomctl.yaml
                           	Darwin:  $HOME/Library/Application Support/bomctl/bomctl.yml
                           	Windows: %AppData%\bomctl\bomctl.yml]
  -v, --verbose            Enable debug output
  ```

**Comment by houdini91** [comment]
@jhoward-lm @idunbarh

I decided to set the default format string to `cyclonedx-1.5`.

* Should we consider changing the default to something else?
* Do we retain the original document format somewhere that the command can use?

This might be related to https://github.com/protobom/protobom/issues/213.

**Review by jhoward-lm — CHANGES_REQUESTED** [review]
Final suggestions to address lint findings and resolve conflicts with `list` PR

**Comment by jhoward-lm** [comment]
> Also for the last [linting issue](https://github.com/bomctl/bomctl/pull/75/files#annotation_23196687002) I think changing the linting rules to allow the max cyclomatic complexity to 13 is reasonable (or take an exception and suppress error).

I think the complexity rule is there for a reason to keep our code easy to read and understand, so I would prefer to avoid modifying the golangci-lint configs if at all possible.

The fix should be pretty simple, just create a private function for the first switch statement, something like

```go
func getFormatMap(encoding string) (map[string]formats.Format, error) {
	switch encoding {
	case formats.JSON:
		return JSONFormatMap(), nil
	case formats.TEXT:
		return TVFormatMap(), nil
	case formats.XML:
		return XMLFormatMap(), nil
	default:
		return formats.EmptyFormat,
			errUnknownEncoding
	}
}
```

jhoward-lm (approved):

cmd/export.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/bomctl/bomctl/internal/pkg/export"
+	"github.com/bomctl/bomctl/internal/pkg/utils"
+	"github.com/bomctl/bomctl/internal/pkg/utils/format"
+)
+
+func exportCmd() *cobra.Command {
+	documentIDs := []string{}
+	opts := &export.ExportOptions{
+		Logger: utils.NewLogger("export"),
+	}
```
jhoward-lm, line 38: Since the logger is instantiated inside `initBackend` right?

```suggestion
	opts := &export.ExportOptions{}
```

cmd/export.go
```diff
@@ -0,0 +1,123 @@
...
+		Args:  cobra.MinimumNArgs(1),
+		Short: "Export SBOM file(s) from Storage",
+		Long:  "Export SBOM file(s) from Storage",
+		PreRun: func(_ *cobra.Command, args []string) {
+			documentIDs = append(documentIDs, args...)
+		},
+		Run: func(cmd *cobra.Command, _ []string) {
+			cfgFile, err := cmd.Flags().GetString("config")
+			cobra.CheckErr(err)
+
```
jhoward-lm, line 55:
```suggestion
			cfgFile, err := cmd.Flags().GetString("config")
			cobra.CheckErr(err)
			
			verbosity, err := cmd.Flags().GetCount("verbose")
			cobra.CheckErr(err)
			
			opts.Debug = verbosity >= minDebugLevel

```

jhoward-lm (approved):
Just some last minute suggestions to leverage the ent client debugging if you get time.

EDIT: these can be done in a follow on issue if time is a factor

houdini91:
@idunbarh @jhoward-lm Squashed offending commit, my bad.

PR #79, opened 2024-06-17
docs: adding usecase examples for updating component metadata and ano…

### LINKED ISSUE #51: docs: describe usecases for bomctl
We're starting to get questions about the differences between GUAC, protobom, and bomctl. We should document the role bomctl plays in this ecosystem of tooling and when it should be used.

We should also encourage people to use these other projects when its better aligned to the user's needs.

idunbarh:
…nymizing sboms

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #51

### DIFF
--- a/docs/usecases/update-primary-component-metadata/bomctl.tar.gz.cdx.json
+++ b/docs/usecases/update-primary-component-metadata/bomctl.tar.gz.cdx.json
@@ -1,0 +1,3857 @@
+{
+  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
+  "bomFormat": "CycloneDX",
+  "specVersion": "1.5",
+  "serialNumber": "urn:uuid:2bc3bc36-ce9b-43fc-83e5-af1efdfc3273",
+  "version": 1,
+  "metadata": {
+    "timestamp": "2024-06-17T19:27:06Z",
+    "tools": {
+      "components": [
+        {
+          "type": "application",
+          "author": "anchore",
+          "name": "syft",
+          "version": "0.103.1"
+        }
+      ]
+    },
+    "component": {
+      "bom-ref": "43bb3ce482554e7f",
+      "type": "file",
+      "name": "bomctl_SNAPSHOT-72b7a56_darwin_amd64.tar.gz",
+      "version": "sha256:e02377eab9eb6ef7edc544a6fe4e76e5d04f4f70720e1eae9441dc1e350fa4d2",
+      "copyright": "bomctl authors",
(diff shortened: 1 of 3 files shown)


docs/usecases/README.md
```diff
@@ -0,0 +1,11 @@
+# Usercases
```
nishakm, line 1: Use Cases

docs/usecases/anonymizing-sboms/README.md
```diff
@@ -0,0 +1,13 @@
+# Anonymizing SBOMs
+
+## Description
+
+Usecase proposed by @mlieberman85 to "anonymize" an sbom (difference from redact) where a bendor is op with putting an SBOM in a platform like GUAC but something goes wrong, and the vendor wants to send a SBOM with sensitive data anonymized but not scrubbed.
```
nishakm, line 5: vendor is ok with?

docs/usecases/anonymizing-sboms/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+So values like `John Smith <john.smith@acme.com>` becomes `Foo Bar <foo.bar@baz.com>`.
```
nishakm, line 7: Oh this is interesting: AFAIK neither one of the SBOM formats support "anonymization" in the sense that both expect a field to be filled or not to be filled (except for SPDX's license NOASSERTION). Wonder if it makes sense to include a "REDACTED" or "ANONYMOUS".

docs/usecases/anonymizing-sboms/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Desire Input
```
nishakm, line 9: Desired

docs/usecases/update-primary-component-metadata/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Desire Input
+
+- [An SBOM of any format](metadata.cyclonedx.json) that only contains the metadata that needs to be updated and associated
+identifier to update.
+- A generated SBOM of any format that comes from syft, trivy, or other tool
+
+## Desired Output
+
+- SBOMs in SPDX and CDX that have the generated SBOM and updated metadata
```
nishakm, line 15: I wonder if a use case here would be creating an "empty" SBOM to be filled out during a review process or something

nishakm (commented):
Just some minor spelling corrections and a couple of ideas.

docs/usecases/anonymizing-sboms/README.md
ashearin:
```suggestion
Usecase proposed by @mlieberman85 to "anonymize" an sbom (difference from redact) where a vendor is ok with putting an SBOM in a platform like GUAC but something goes wrong, and the vendor wants to send a SBOM with sensitive data anonymized but not scrubbed.
```

PR #103, opened 2024-07-11
docs: ADR-0003: customization of list output

jhoward-lm:
# 3. List output customization

Date: 2024-07-11

## Status

Proposed

## Context

This is an enhancement proposal to allow users to customize the output of `bomctl list` by specifying
a search expression, filter, or template.

## Decision

Update `bomctl list` to accept an expression to customize the output table. This would be similar in
functionality to the `--format` option used by various `docker` commands.

Either the existing use of positional argument will be retained, or an additional flag (such as
`--format` or `--query`) will be added.

The supported search/filter expression syntaxes will be:

- Common Expression Language (CEL)
- Go templates
- GraphQL

The expression would then be parsed to construct a SQL query or call to the `ent` backend to
retrieve and display the desired table(s)/column(s).

## Consequences

This change would allow users greater control over inspecting the contents of their cached SBOMs.

### DIFF
--- a/docs/architecture/0003-list-output-customization.md
+++ b/docs/architecture/0003-list-output-customization.md
@@ -1,0 +1,54 @@
+<!--
+This is a template for [Documenting Architecture Decisions - Michael Nygard](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions).
+
+You can use [adr-tools](https://github.com/npryce/adr-tools) for managing the ADR files.
+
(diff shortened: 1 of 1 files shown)


eddiezane:
https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/cli-runtime/pkg/printers

ashearin (approved):

PR #104, opened 2024-07-11
docs: ADR-0004: authentication types and encrypted secrets

jhoward-lm:
# 4. Authentication types and secrets encryption

Date: 2024-07-11

## Status

Proposed

## Context

Additional authentication mechanisms and secure secrets storage should be supported.

## Decision

Add support for additional authentication mechanisms:

- OAuth
- Bearer token
- Basic

Secrets provided as plain text will be encrypted with a user-provided or auto-generated key pair.

Add a config file mapping of URLs or bare hostnames to associated user credentials.

- If credentials are specified by the user or encountered in the config file as plain text:
  - If no private key is provided by user or already exists in the config directory,
    auto-generate a new key pair
  - Encrypt secrets inline in the config file using the private key

Either leverage the [SOPS](https://getsops.io) tool as a library to perform the encryption/decryption,
or use its encrypted string expression form.

Example of proposed config file additions:

```yaml
auths:
  github.com: ENC[AES256_GCM,data:Tr7o=,iv:1=,aad:No=,tag:k=]
  gitlab.com:
    user: ENC[AES256_GCM,data:CwE4O1s=,iv:2k=,aad:o=,tag:w==]
    password: ENC[AES256_GCM,data:p673w==,iv:YY=,aad:UQ=,tag:A=]
```

## Consequences

These changes will increase flexibility for users by allowing fetching from and pushing to additional
remote endpoints that may have limited or strict options for authentication. They will also provide
enhanced security options for local storage and transmission of secrets.

### DIFF
--- a/docs/architecture/0004-authentication-types-and-secrets-encryption.md
+++ b/docs/architecture/0004-authentication-types-and-secrets-encryption.md
@@ -1,0 +1,68 @@
+<!--
+This is a template for [Documenting Architecture Decisions - Michael Nygard](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions).
+
+You can use [adr-tools](https://github.com/npryce/adr-tools) for managing the ADR files.
+
+In each ADR file, write the following sections.
+-->
+# 4. Authentication types and secrets encryption
+
+Date: 2024-07-11
+
+## Status
+<!--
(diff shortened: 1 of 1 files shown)


ashearin:
Per working group session 8/5/24. For now leave as proposed, be reevaluated for a future release.

ashearin (approved):

lmphil (approved):

PR #105, opened 2024-07-11
docs: ADR-0005: SBOM document aliases/tags

jhoward-lm:
# 5. SBOM document aliases/tags

Date: 2024-07-11

## Status

Proposed

## Context

A shorter form of referencing SBOM documents stored in `bomctl`'s database is needed.

## Decision

Add support for user assignment of one of the following options per stored SBOM document:

- a single alias
- a list of one or more tags

These options could also be consolidated into a unified concept such as **labels**.

Users should be able to perform this action either at fetch/import or as a standalone command/operation.

## Consequences

Implementing aliases could be limiting due to their 1:1 association.

Using the tags methodology may provide greater flexibility in association between documents for actions like storing and retrieving a group of SBOMs in a tree structure.

In addition, tags can still be utilized to perform a function similar to what aliases provide.

### DIFF
--- a/docs/architecture/0005-sbom-document-aliases-tags.md
+++ b/docs/architecture/0005-sbom-document-aliases-tags.md
@@ -1,0 +1,52 @@
+<!--
+This is a template for [Documenting Architecture Decisions - Michael Nygard](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions).
+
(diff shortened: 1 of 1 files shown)


mfrystacky:
Having `alias` and a separate `tags` concept makes a lot of sense.

ashearin (approved):

PR #111, opened 2024-07-29
docs: ADR-0006 sbom merge

pkwiatkowski1:
# 6. Merge subcommand

Date: 2024-07-29

## Status

Accepted

## Context

SBOM merge capability should be supported.

## Decision

- Ability to take one or more SBOMs in different formats into a singular sbom
  - Merge components that may match on different identifiers (purls, hash, etc …)
    - Merge component dependencies
    - Merge component properties

- Ability to take a flag to “flatten” any externally referenced SBOMs
  - Any nested components should be moved to "top level" components and the depends_on section updated to show the dependency.

## Consequences

These changes allow end users to easily consolidate multiple SBOMs from a variety of different sources into one while still being agnostic of the SBOM format being used.


docs/architecture/0006-sbom-merge.md
```diff
@@ -0,0 +1,47 @@
+<!--
+This is a template for [Documenting Architecture Decisions - Michael Nygard](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions).
+
+You can use [adr-tools](https://github.com/npryce/adr-tools) for managing the ADR files.
+
+In each ADR file, write the following sections.
+-->
+# 6. 0006-sbom-merge.md
```
jhoward-lm, line 8:
```suggestion
# 6. merge subcommand
```

jhoward-lm (changes requested):

jhoward-lm:
Just to document here for whoever ends up working on this, I think an implementation using a Go generic interface would be a clean way to do it.

<details>

<summary>Basic example</summary>

```go
package main

import (
	"fmt"

	"github.com/protobom/protobom/pkg/sbom"
)

type (
	Merger[T any] interface {
		Merge(*T) error
	}

	MergerBase[T any] struct {
		Base *T
		Merger[T]
	}
)

func (*MergerBase[T]) Merge(t *T) error {
	switch any(t).(type) {
	case *sbom.Metadata:
		fmt.Println("Merging metadata")
	case *sbom.Node:
		if _, ok := any(m.Base).(*sbom.Node); !ok {
			// return error
		}

		fmt.Println("Merging nodes")
	case *sbom.NodeList:
		fmt.Println("Merging node lists")
	case *sbom.Person:
		fmt.Println("Merging persons")
	case *sbom.Tool:
		fmt.Println("Merging tools")
	default:
		// return error
	}

	return nil
}

// Enforce implementation of interface at compile time
var _ Merger[sbom.Node] = (*MergerBase[sbom.Node])(nil)

func main() {
	merger := &MergerBase[sbom.Node]{}
	merger.Merge(&sbom.Node{})
}
```

</details>

[Go playground link](https://go.dev/play/p/-ApK7LFy0xv)

ashearin (approved):

jhoward-lm (approved):

PR #112, opened 2024-07-30
docs: ADR-0007 sbom diff

mfrystacky:
## 7. SBOM Diff

Date: 2024-07-30
## Status

Proposed
## Context

Support the ability to see the difference between two SBOMs.
## Decision

Seeing where two SBOMs differ and highlighting potentially problematic changes such as:

    a dependency that is slightly renamed and pointing to a different url
    a new dependency that is pointing to a new/unknown package
    a dependency whose version is bumped down

Some tools do that do that on the package lock level since that sort of supply chain attack is starting to get a bit more prevalent.

Are there other useful things we should highlight for the user?
## Consequences


jhoward-lm:
Mind updating the PR description like the others with the content of the ADR document?

mfrystacky:
Done!

jhoward-lm:
> Done!

I meant copy the content of the new ADR document, edit the PR description, and replace everything with paste

djmoch:
Capturing the thoughts that I brought up in the working meeting:

- We should make clear that the default output will output _all_ of the changes. Whatever we think we need to do to highlight information we think is critical is fine, but I do think users will expect to see all changes given the `diff` verb.
- Similarly, I agree with everyone else in the meeting that the most obvious output _format_ would be a textual, `git diff`-style output. I'm not persuaded it's wise to invest in other output formats until a stronger culture develops around how to represent changes between SBOM's.

jhoward-lm (approved):

ashearin (approved):

Issue #46, opened 2024-05-02
Push Subcommand


mfrystacky:
Add push command capability to bomctl

### Desired functionality

Ability to upload an SBOM or a series of SBOMs to external hosts via
 - OCI
 - PUT/POST HTTPS/2 
 - Git commit to repo
 - File systems

PR #122, opened 2024-08-08
feat: add push subcommand cli

ashearin:
# Pull Request Template

## Description

Adds base cli cmd and options for push subcommand. As well as base scaffolding in pkg/internal to begin adding client functionality. 

Future PRs would add client functionality for oci/http/git, etc

## Type of change

## How Has This Been Tested?

No unit tests added yet

Related to #46

### DIFF
--- a/cmd/push.go
+++ b/cmd/push.go
@@ -35,13 +35,13 @@ func pushCmd() *cobra.Command {
 	}
 
 	pushCmd := &cobra.Command{
-		Use:    "push [flags] SBOM_ID DEST_PATH...",
+		Use:    "push [flags] SBOM_ID [SBOM_ID...] DEST_PATH",
 		Args:   cobra.MinimumNArgs(pushArgNum),
 		Short:  "Push stored SBOM file(s) to remote URL or filesystem",
 		Long:   "Push stored SBOM file(s) to remote URL or filesystem",
 		PreRun: preRun(opts.Options),
 		Run: func(_ *cobra.Command, args []string) {
-			if err := push.Push(args[0], args[1], opts); err != nil {
+			if err := push.Push(args[:len(args)-1], args[len(args)-1], opts); err != nil {
 				opts.Logger.Fatal(err)
 			}
 		},
--- a/cmd/push.go
+++ b/cmd/push.go
@@ -1,0 +1,56 @@
+// ------------------------------------------------------------------------
+// SPDX-FileCopyrightText: Copyright © 2024 bomctl a Series of LF Projects, LLC
+// SPDX-FileName: cmd/push.go
+// SPDX-FileType: SOURCE
+// SPDX-License-Identifier: Apache-2.0
+// ------------------------------------------------------------------------
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


cmd/push.go
```diff
@@ -0,0 +1,56 @@
...
+			if err := push.Push(args[0], args[1], opts); err != nil {
+				opts.Logger.Fatal(err)
+			}
+		},
+	}
+
+	pushCmd.Flags().StringP("encoding", "e", formats.JSON, encodingHelp())
+	pushCmd.Flags().StringP("format", "f", formats.CDXFORMAT, formatHelp())
+	pushCmd.Flags().BoolVar(&opts.UseNetRC, "netrc", false, "Use .netrc file for authentication to remote hosts")
+	pushCmd.Flags().BoolVar(&opts.UseTree, "tree", false, "Push all Sboms in Tree to specified Destination")
```
jhoward-lm, line 53:
```suggestion
	pushCmd.Flags().BoolVar(&opts.UseTree, "tree", false, "Recursively push all SBOMs in external reference tree")
```

cmd/push.go
```diff
@@ -0,0 +1,56 @@
...
+func pushCmd() *cobra.Command {
+	opts := &push.Options{
+		Options: options.New(options.WithLogger(utils.NewLogger("push"))),
+	}
+
+	pushCmd := &cobra.Command{
+		Use:    "push [flags] SBOM_ID DEST_PATH...",
+		Args:   cobra.MinimumNArgs(pushArgNum),
+		Short:  "Push SBOM file(s) to HTTP(S) URLs, OCI, Git Repos, FileSystem",
+		Long:   "Push SBOM file(s) to HTTP(S) URLs, OCI, Git Repos, FileSystem",
```
jhoward-lm, line 41:
```suggestion
		Short:  "Push stored SBOM file(s) to remote URL or filesystem",
		Long:   "Push stored SBOM file(s) to remote URL or filesystem",
```

internal/pkg/push/push.go
```diff
@@ -0,0 +1,130 @@
...
+	document, err := backend.GetDocumentByID(sbomID)
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+
+	// Push externally referenced BOMs
+	return pushDocument(document, destPath, opts)
+}
+
+func NewPusher(sbomURL string) (Pusher, error) {
```
jhoward-lm, line 75: Probably should have done this with fetcher as well, so it's called from outside this package as `pusher.New()` instead of `pusher.NewPusher()`

```suggestion
func New(sbomURL string) (Pusher, error) {
```

cmd/push.go
```diff
@@ -0,0 +1,56 @@
...
+
+const pushArgNum = 2
+
+func pushCmd() *cobra.Command {
+	opts := &push.Options{
+		Options: options.New(options.WithLogger(utils.NewLogger("push"))),
+	}
+
+	pushCmd := &cobra.Command{
+		Use:    "push [flags] SBOM_ID DEST_PATH...",
```
jhoward-lm, line 38: I think if using positional args this should be more like `push [flags] SBOM_ID [SBOM_ID...] DEST_PATH` (push one or more SBOM IDs to single destination)

Then below might look something like

```go
			if err := push.Push(args[:len(args)-1], args[len(args)-1], opts); err != nil {
				opts.Logger.Fatal(err)
			}
```

ashearin: I wasn't sure how we would handle multiple sboms, I think it's reasonable to limit the destination to one and let them push as many as they want to one spot

ashearin: Updated to handle multiple ids

jhoward-lm (approved):

PR #134, opened 2024-08-16
fix: recursive fetching and outputting of recursive files

### LINKED ISSUE #133: `bomctl fetch` not recursively fetch
The following command should collect two sboms.

```
bomctl fetch https://raw.githubusercontent.com/bomctl/bomctl-playground/main/examples/bomctl-container-image/bomctl_bomctl_v0.3.0.cdx.json
```

but recursive fetching is not occurring due to [this line](https://github.com/bomctl/bomctl/blob/v0.3.0/internal/pkg/fetch/fetch.go#L81).

If you run the same command but add an output option it will recursively run. I think this behavior is the opposite if what is expected.

idunbarh:
## Description

This PR fixes three issues.

- Recursive fetching of SBOMs was not occurring be unless an output file was specified
- The recursive output files had an exta "." in the name
- When recursive fetching and a output file was specified, the recursive output files were empty

Fixes #133 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

Ran ...
```
bomctl fetch https://raw.githubusercontent.com/bomctl/bomctl-playground/main/examples/bomctl-container-image/bomctl_bomctl_v0.3.0.cdx.json
```
... with and without an output file.

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings

### DIFF
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -42,24 +42,22 @@ func fetchCmd() *cobra.Command {
 		Long:   "Fetch SBOM file(s) from HTTP(S), OCI, or Git URLs",
 		PreRun: preRun(opts.Options),
 		Run: func(_ *cobra.Command, args []string) {
-			var outputFile *os.File
-
-			if string(outputFileName) != "" {
+			if outputFileName != "" {
 				if len(args) > 1 {
 					opts.Logger.Fatal("The --output-file option cannot be used when more than one URL is provided.")
 				}
 
 				out, err := os.Create(string(outputFileName))
 				if err != nil {
-					opts.Logger.Fatal("error creating output file", "outputFile", outputFile)
+					opts.Logger.Fatal("error creating output file", "outputFileName", outputFileName)
 				}
 
-				outputFile = out
-				defer outputFile.Close()
+				opts.OutputFile = out
+				defer opts.OutputFile.Close()
 			}
 
 			for _, url := range args {
-				if err := fetch.Fetch(url, opts, outputFile); err != nil {
+				if err := fetch.Fetch(url, opts); err != nil {
 					opts.Logger.Fatal(err)
 				}
 			}
--- a/internal/pkg/fetch/fetch.go
+++ b/internal/pkg/fetch/fetch.go
@@ -36,8 +36,8 @@ import (
 	"github.com/bomctl/bomctl/internal/pkg/options"
 )
 
-func Fetch(sbomURL string, opts *options.FetchOptions, outputFile *os.File) error {
-	document, err := fetchDocument(sbomURL, opts, outputFile)
+func Fetch(sbomURL string, opts *options.FetchOptions) error {
+	document, err := fetchDocument(sbomURL, opts)
 	if err != nil {
 		return err
 	}
@@ -57,10 +57,10 @@ func Fetch(sbomURL string, opts *options.FetchOptions, outputFile *os.File) erro
 	}
 
 	// Fetch externally referenced BOMs
-	return fetchExternalReferences(document, backend, opts, outputFile)
+	return fetchExternalReferences(document, backend, opts)
 }
 
-func fetchDocument(sbomURL string, opts *options.FetchOptions, outputFile *os.File) (*sbom.Document, error) {
+func fetchDocument(sbomURL string, opts *options.FetchOptions) (*sbom.Document, error) {
 	fetcher, err := client.New(sbomURL)
 	if err != nil {
 		return nil, fmt.Errorf("creating fetch client: %w", err)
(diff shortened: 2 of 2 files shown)


cmd/fetch.go
```diff
@@ -42,30 +42,31 @@ func fetchCmd() *cobra.Command {
 		Long:   "Fetch SBOM file(s) from HTTP(S), OCI, or Git URLs",
 		PreRun: preRun(opts.Options),
 		Run: func(_ *cobra.Command, args []string) {
-			if string(outputFile) != "" {
+			var outputFile *os.File
+
+			if string(outputFileName) != "" {
```
jhoward-lm, line 47:
```suggestion
			if outputFileName != "" {
```

cmd/fetch.go
```diff
@@ -42,30 +42,31 @@ func fetchCmd() *cobra.Command {
...
-				opts.OutputFile = out
-
-				defer opts.OutputFile.Close()
+				outputFile = out
+				defer outputFile.Close()
 			}
 
 			for _, url := range args {
-				if err := fetch.Fetch(url, opts); err != nil {
+				if err := fetch.Fetch(url, opts, outputFile); err != nil {
```
jhoward-lm, line 62: I think the variable name change is a good idea and makes the code clearer.

Why are we bypassing the `opts.OutputFile` though? The goal of using the options struct is so that the parameter sets of entry point functions like `Fetch` don't change.

Since this PR doesn't remove `OutputFile` from the `FetchOptions` struct, was this perhaps unintentional?

internal/pkg/fetch/fetch.go
```diff
@@ -36,8 +36,8 @@ import (
 	"github.com/bomctl/bomctl/internal/pkg/options"
 )
 
-func Fetch(sbomURL string, opts *options.FetchOptions) error {
-	document, err := fetchDocument(sbomURL, opts)
+func Fetch(sbomURL string, opts *options.FetchOptions, outputFile *os.File) error {
+	document, err := fetchDocument(sbomURL, opts, outputFile)
```
jhoward-lm, line 40: I think this should remain how it was, referencing `opts.OutputFile` and without the unneeded `outputFile` param.

internal/pkg/fetch/fetch.go
```diff
@@ -56,15 +56,11 @@ func Fetch(sbomURL string, opts *options.FetchOptions) error {
 		return fmt.Errorf("error adding document: %w", err)
 	}
 
-	if opts.OutputFile == nil || opts.OutputFile.Name() == "" {
-		return nil
-	}
-
 	// Fetch externally referenced BOMs
-	return fetchExternalReferences(document, backend, opts)
+	return fetchExternalReferences(document, backend, opts, outputFile)
```
jhoward-lm, line 60: I think this should remain how it was, referencing `opts.OutputFile` and without the unneeded `outputFile` param.

internal/pkg/fetch/fetch.go
```diff
@@ -56,15 +56,11 @@ func Fetch(sbomURL string, opts *options.FetchOptions) error {
...
 }
 
-func fetchDocument(sbomURL string, opts *options.FetchOptions) (*sbom.Document, error) {
+func fetchDocument(sbomURL string, opts *options.FetchOptions, outputFile *os.File) (*sbom.Document, error) {
```
jhoward-lm, line 63: I think this should remain how it was, referencing `opts.OutputFile` and without the unneeded `outputFile` param.

internal/pkg/fetch/fetch.go
```diff
@@ -94,21 +90,26 @@ func fetchDocument(sbomURL string, opts *options.FetchOptions) (*sbom.Document,
 	return document, nil
 }
 
-func fetchExternalReferences(document *sbom.Document, backend *db.Backend, opts *options.FetchOptions) error {
+func fetchExternalReferences(document *sbom.Document, backend *db.Backend,
+	opts *options.FetchOptions, outputFile *os.File,
+) error {
```
jhoward-lm, line 95: I think this should remain how it was, referencing `opts.OutputFile` and without the unneeded `outputFile` param.

jhoward-lm (changes requested):
Minor comments about reverting to using `opts.OutputFile` instead of additional `outputFile *os.File` parameters in multiple places, but otherwise looks great!

cmd/fetch.go
jhoward-lm: I'll have to give it some more thought, but I think what makes the most sense to me would be creating a new instance of `FetchOptions` with the new `OutputFile` on it to keep the function signature the same.

Basically recreate the setup process of when `Fetch` was originally called from `cmd/fetch.go`.

ashearin (approved):

internal/pkg/fetch/fetch.go
```diff
@@ -101,14 +97,18 @@ func fetchExternalReferences(document *sbom.Document, backend *db.Backend, opts
 	}
 
 	for _, ref := range extRefs {
-		refOutput, err := getRefFile(opts.OutputFile)
-		if err != nil {
-			return err
+		extRefsOpt := *opts
```
jhoward-lm, line 100: Is this actually doing a copy by value, or are the two variables still referencing the same underlying memory address?

What does it say if, for debugging purposes, you add this before and after the call to `getRefFile()`:

```go
fmt.Printf("opts: %p\n", opts.OutputFile)
fmt.Printf("extRefsOpt: %p\n", extRefsOpt.OutputFile)
```

jhoward-lm (approved):

PR #138, opened 2024-08-22
refactor: dynamic prefix for subcommand loggers

jhoward-lm:
## Description

This PR includes a small refactor to use a dynamic prefix when creating a subcommand's logger, and only call the logger instantiation from a single location in `rootCmd` instead of from every subcommand.

The intent is to use the first top-level subcommand's name in cases where a subcommand may have more than one nested subcommand (e.g. `bomctl alias remove ...`, in which case the logger prefix will be `alias:` instead of `remove:`).

## Type of change

- [X] Refactor

## Checklist

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -33,7 +33,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/bomctl/bomctl/internal/pkg/export"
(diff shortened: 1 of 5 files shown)


ashearin (approved):
Awesome!

PR #152, opened 2024-09-07
chore: enable protogetter and varnamelen linters

jhoward-lm:
## Description

This PR enables the `protogetter` and `varnamelen` linters for `golanci-lint`.

### protogetter

Reports direct reads from proto message fields when getters should be used.

This should significantly reduce instances of panic due to `invalid memory address or nil pointer dereference`. Fixes were automatically applied with `golangci-lint run --fix`, except lines 230-243 in merger.go which generated incorrect code and needed manual edits.

### varnamelen

Checks that the length of a variable's name matches its scope.

Fixes were manually applied, but no logic was modified.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Verified `task lint` and `task test` locally.

## Checklist

### DIFF
--- a/internal/pkg/merge/merger.go
+++ b/internal/pkg/merge/merger.go
@@ -43,22 +43,22 @@ var (
 	errCastFailure  = errors.New("failed to cast object")
 )
 
-func (m *MergerBase[T]) MergeProtoMessage(t T) (err error) {
+func (m *MergerBase[T]) MergeProtoMessage(msg T) (err error) {
 	switch base := any(m.Base).(type) {
 	case *sbom.Metadata:
-		return mergeMetadata(base, t)
+		return mergeMetadata(base, msg)
 	case *sbom.Node:
-		return mergeNode(base, t)
+		return mergeNode(base, msg)
 	case *sbom.NodeList:
-		return mergeNodeList(base, t)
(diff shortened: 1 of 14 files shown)


internal/pkg/merge/merger.go
```diff
@@ -227,16 +227,19 @@ func mergeDocumentType(base *sbom.DocumentType, t proto.Message) error {
 		return fmt.Errorf("%w: %T", errCastFailure, t)
 	}
 
-	if (base.Name == nil || *base.Name == "") && other.Name != nil {
-		base.Name = other.Name
+	if base.GetName() == "" && other.GetName() != "" {
+		name := other.GetName()
+		base.Name = &name
```
ashearin, line 232: I'm assuming not since you didn't do it, but might as well ask:
Can we not do?
```suggestion
		base.Name = &other.GetName()
```

ashearin (approved):
One question a section of changes, but overall gtg

internal/pkg/merge/merger.go
jhoward-lm: Yeah I did try that. Go doesn't let you address a value, only a variable. I assume it would be similar to trying to write `&"some string"`

PR #153, opened 2024-09-11
feat: document annotations for source data, hash, format, and URL

### LINKED ISSUE #68: Extended ent schemas
Specifically for things like mapping document ID to raw SBOM byte data

jhoward-lm:
## Description

This PR adds annotations in the database for each document fetched. The new annotation names are:

- `bomctl_annotation_source_data` - bytes content of fetched document
- `bomctl_annotation_source_hash` - hash of fetched document
- `bomctl_annotation_source_format` - original format of fetched document
- `bomctl_annotation_source_url` - URL of fetched document 

Fixes #68 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [X] updated Go tests
- [X] run locally and verify database contents

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -84,7 +84,7 @@
   "go.coverMode": "atomic",
   "go.coverOnSave": true,
   "go.coverOnSingleTest": true,
-  "go.lintFlags": ["--fast", "--allow-parallel-runners", "--config=.golangci.yml"],
+  "go.lintFlags": ["--fast"],
   "go.lintOnSave": "file",
   "go.lintTool": "golangci-lint",
   "go.testFlags": ["-race", "-covermode=atomic"],
--- a/cmd/fetch.go
+++ b/cmd/fetch.go
@@ -58,7 +58,7 @@ func fetchCmd() *cobra.Command {
 			}
 
 			for _, url := range args {
-				if err := fetch.Fetch(url, opts); err != nil {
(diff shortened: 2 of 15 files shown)


ashearin (approved):
Looks good!


One question about this section in fetch: (69-75)
```python
if err := backend.SetUniqueAnnotation(
		doc.GetMetadata().GetId(), db.SourceURLAnnotation, sbomURL,
	); err != nil {
		return nil, fmt.Errorf(
			"applying unique annotation %s to %s: %w", db.SourceURLAnnotation, doc.GetMetadata().GetId(), err,
		)
	}
```

Formatting is a little weird, but we're setting the source url as the id in the annotations? Might have that backwards, but are we using this anywhere currently? If not, whats the plan

jhoward-lm:
> we're setting the source url as the id in the annotations? Might have that backwards, but are we using this anywhere currently? If not, whats the plan

So, the signature is `SetUniqueAnnotation(documentID string, name string, value string) error`. In other words, "on the document with `documentID`, set a unique annotation named `name` with a value of `value`".

Not currently using it anywhere, but it enables the capability to look up a document in the database by its URL if it was already fetched. One use case might be if we're trying to fetch a whole tree we can check the database first to see if the external reference was already fetched, and if so, use it.

lmphil (approved):

PR #165, opened 2024-09-18
fix: Print DocID for list cmd. Dynamic id/alias column width

### LINKED ISSUE #162: BUG: SPDX IDs from bomctl list not usable
Currently we're using the document "name"  as the ID for SPDX documents for `bomctl list` while for cylonedx we're using the ID (UUID from the cdx sbom)

Example:
```bash
bomctl-playground (main) $ bomctl fetch https://raw.githubusercontent.com/bomctl/bomctl-playground/main/examples/bomctl-container-image/bomctl_bomctl_v0.3.0.cdx.json
INFO  fetch: Fetching from HTTP URL url=https://raw.githubusercontent.com/bomctl/bomctl-playground/main/examples/bomctl-container-image/bomctl_bomctl_v0.3.0.cdx.json
INFO  fetch: Fetching from HTTP URL url=https://raw.githubusercontent.com/bomctl/bomctl-playground/main/examples/bomctl-container-image/app/bomctl_0.3.0_linux_amd64.tar.gz.spdx.json

bomctl-playground (main) $ bomctl list

                        ID                        │ Version  │ # Nodes  
──────────────────────────────────────────────────┼──────────┼──────────
 urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db    │    1     │    6     
 bomctl_0.3.0_linux_amd64.tar.gz                  │    0     │    81  
```

This creates an issue when attempting to use the SPDX SBOM in future commands, as the name is not a valid document in the db: 

```bash
bomctl-playground (main) $ bomctl export bomctl_0.3.0_linux_amd64.tar.gz -o test.spdx.json
INFO  export: Exporting document sbomID=bomctl_0.3.0_linux_amd64.tar.gz
WARN  db: Document could not be retrieved id=bomctl_0.3.0_linux_amd64.tar.gz err="no documents matching IDs bomctl_0.3.0_linux_amd64.tar.gz"
FATAL export: failed to retrieve document: no documents matching IDs bomctl_0.3.0_linux_amd64.tar.gz
```

We should be outputting the correct database ID for each document as part of `bomctl list`.

* There is a larger discussion to be had on how we standardize the ID used within bomctl for documents, since there is a disparity on how the major formats handle this. CDX provides a uuid whereas SPDX simply uses the filename. Protobom currently tacks on some additional info to the spdx document name to use as the id internally, which is what `bomctl list` should be showing so thats its a usable string in bomctl operations. A separate issue/ADR will be opened to discuss this solution.

ashearin:
## Description

Fixes spdx SBOMs not getting full cache ID printed in `list` cmd. 

Dynamically sets column width for ID and Alias contents. version and # nodes are still static widths

Determines terminal width and total table width needed, prints out a list of SBOMs if the width of terminal is smaller than width of table. otherwise prints out the table. 

Fixes #162

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Output when width of table is less than width of terminal:
```bash
$ go run . list

                                                     ID                                                      │    Alias     │ Version │ # Nodes 
─────────────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────┼─────────┼─────────
 urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db                                                               │ bomctl_0.3.0 │    1    │    6    
 https://anchore.com/syft/file/bomctl_0.3.0_linux_amd64.tar.gz-1b838d44-9d3c-47d0-9f7f-846397e701fa#DOCUMENT │              │    0    │   81    
```

Output if terminal width is less than table width:
```bash
$ go run . list
ID: urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db
Alias: bomctl_0.3.0
Version: 1
# Nodes: 6

ID: https://anchore.com/syft/file/bomctl_0.3.0_linux_amd64.tar.gz-1b838d44-9d3c-47d0-9f7f-846397e701fa#DOCUMENT
Alias: 
Version: 0
# Nodes: 81
```
## Checklist

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -22,221 +22,11 @@ package cmd
 import (
 	"fmt"
 	"os"
-	"strings"
 
-	"github.com/charmbracelet/lipgloss"
-	lgtable "github.com/charmbracelet/lipgloss/table"
-	"github.com/protobom/protobom/pkg/sbom"
 	"github.com/spf13/cobra"
-	"golang.org/x/term"
 
-	"github.com/bomctl/bomctl/internal/pkg/db"
+	"github.com/bomctl/bomctl/internal/pkg/format"
 )
-
-const (
-	columnIdxID = iota
-	columnIdxAlias
-	columnIdxVersion
-	columnIdxNumNodes
-
-	columnNameID       = "ID"
-	columnNameAlias    = "Alias"
-	columnNameVersion  = "Version"
-	columnNameNumNodes = "# Nodes"
-
-	columnWidthID       = 47
-	columnWidthAlias    = 12
-	columnWidthVersion  = 9
-	columnWidthNumNodes = 9
-
-	cellSideCount = 2
-
-	paddingHorizontal = 1
-	paddingVertical   = 0
-
-	rowHeaderIdx = 0
-	rowMaxHeight = 1
-
-	totalColumnCount = 4
-)
-
-type (
-	columnDefinition struct {
-		name  string
-		width int
-	}
-
-	rowData struct {
-		id, alias, version, numNodes string
-	}
-
-	table struct {
-		columns []columnDefinition
-		rows    []rowData
-	}
-)
-
-func (t *table) String() string {
-	tooWide := t.determineColumnWidths()
-
-	if tooWide {
-		return formatList(t)
-	}
-
-	return formatTable(t)
-}
-
-func newTable() *table {
-	cols := []columnDefinition{}
-
-	for c := range totalColumnCount {
-		name := ""
-		width := 0
-
-		switch c {
-		case columnIdxID:
-			name = columnNameID
-			width = columnWidthID
-		case columnIdxAlias:
-			name = columnNameAlias
-			width = columnWidthAlias
-		case columnIdxVersion:
-			name = columnNameVersion
-			width = columnWidthVersion
-		case columnIdxNumNodes:
-			name = columnNameNumNodes
-			width = columnWidthNumNodes
-		}
-
-		cols = append(cols, columnDefinition{
-			name:  name,
-			width: width,
-		})
-	}
-
-	return &table{
-		columns: cols,
-		rows:    []rowData{},
-	}
-}
-
-func getTermInfo() int {
-	width, _, err := term.GetSize(0)
-	if err != nil {
-		return 0
-	}
-
-	return width
-}
-
-func (t *table) getHeaders() []string {
-	headers := []string{}
-
-	for c := range totalColumnCount {
-		headers = append(headers, t.columns[c].name)
-	}
-
-	return headers
-}
-
-func (t *table) getRows() [][]string {
-	rows := [][]string{}
-
-	for _, row := range t.rows {
-		rows = append(rows, []string{row.id, row.alias, row.version, row.numNodes})
-	}
-
-	return rows
-}
-
-func (t *table) getTableWidth() int {
-	totalWidth := totalColumnCount * (paddingHorizontal * cellSideCount)
-
-	for c := range totalColumnCount {
-		totalWidth += t.columns[c].width
-	}
-
-	return totalWidth
-}
-
-func (t *table) determineColumnWidths() bool {
-	terminalWidth := getTermInfo()
-	padding := paddingHorizontal * cellSideCount
-
-	for _, row := range t.rows {
-		if idWidth := len(row.id); idWidth >= t.columns[columnIdxID].width {
-			t.columns[columnIdxID].width = idWidth + padding
-		}
-
-		if aliasWidth := len(row.alias); aliasWidth >= t.columns[columnIdxAlias].width {
-			t.columns[columnIdxAlias].width = aliasWidth + padding
-		}
-
-		if versionWidth := len(row.version); versionWidth >= columnIdxVersion {
-			t.columns[columnIdxVersion].width = versionWidth + padding
-		}
-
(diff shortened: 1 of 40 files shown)


cmd/list.go
```diff
@@ -44,13 +45,90 @@ const (
...
+	totalWidth := (idColumnWidth + aliasColumnWidth + columnWidthVersion + columnWidthNumNodes)
+	totalWidth += headerCount * (paddingHorizontal * cellSideCount)
+
+	return (totalWidth > terminalWidth), []int{idColumnWidth, aliasColumnWidth, columnWidthVersion, columnWidthNumNodes}
+}
+
+func printList(rows ...[]string) {
+	for _, row := range rows {
+		fmt.Fprintf(os.Stdout, "ID: %s\n"+"Alias: %s\n"+"Version: %s\n"+"# Nodes: %s\n\n",
+			row[columnIdxID], row[columnIdxAlias], row[columnIdxVersion], row[columnIdxNumNodes])
```
jhoward-lm, line 90: What do you think about either:

```go
		fmt.Fprintf(os.Stdout,
			strings.Join([]string{
				"ID:      %s",
				"Alias:   %s",
				"Version: %s",
				"# Nodes: %s",
				"", "",
			}, "\n"),
			row[columnIdxID], row[columnIdxAlias], row[columnIdxVersion], row[columnIdxNumNodes],
		)
```

or:

```go
		fmt.Fprintf(os.Stdout, "%-8s: %s\n%-8s: %s\n%-8s: %s\n%-8s: %s\n\n",
			"ID", row[columnIdxID],
			"Alias", row[columnIdxAlias],
			"Version", row[columnIdxVersion],
			"# Nodes", row[columnIdxNumNodes],
		)
```

cmd/list.go
```diff
@@ -44,13 +45,90 @@ const (
...
+	}
+}
```
jhoward-lm, line 92: I don't think the print should happen in the function directly, it should just return a formatted string to print from the calling context

```suggestion
func formatList(rows ...[]string) string {
	output := ""

	for _, row := range rows {
		output = strings.Join([]string{
			output,
			fmt.Sprintf("ID      : %s", row[columnIdxID]),
			fmt.Sprintf("Alias   : %s", row[columnIdxAlias]),
			fmt.Sprintf("Version : %s", row[columnIdxVersion]),
			fmt.Sprintf("# Nodes : %s", row[columnIdxNumNodes]),
			"",
		}, "\n")
	}

	return output
}
```

and call it with `fmt.Fprint(os.Stdout, formatList(rows...))`

cmd/list.go
```diff
@@ -44,13 +45,90 @@ const (
...
+
+func printTable(colWidths []int, rows ...[]string) {
```
jhoward-lm, line 94: Same comment as above about printing directly in the called function, rename to `formatTable` and return a string to print from the calling context

cmd/list.go
jhoward-lm, line 1: Management of row data is getting a little complex. I think we should add

- a type to hold a column definition
- a type to hold row data
- a type to hold all table data

Something along the lines of:

```go
type (
	columnDefinition struct {
		name  string
		width int
	}

	rowData struct {
		id, alias, version, numNodes string
	}

	table struct {
		columns []columnDefinition
		rows    []rowData
	}
)

func(t *table) String() string {
	output := ""

	for _, row := range t.rows {
		output = strings.Join([]string{
			output,
			fmt.Sprintf("ID      : %s", row.id),
			fmt.Sprintf("Alias   : %s", row.alias),
			fmt.Sprintf("Version : %s", row.version),
			fmt.Sprintf("# Nodes : %s", row.numNodes),
			"",
		}, "\n")
	}

	return output 
}
```

ashearin: I tried to do something similar with lipgloss types and I couldn't get it to work the way we needed. I'll work it from this direction

internal/pkg/format/table.go
jhoward-lm, line 1: I think the package name `format` might be a little confusing since we typically are talking about SBOM formats when we use that word, and sort of blurs the lines between the `protobom` `formats` package and this one.

What about something like`infoutils` with the table and list logic split into two files?

```
infoutils/
├── list.go
└── table.go
```

Other package names I thought of but not sure if any of them make sense

- `presentation`
- `outpututils`
- `dbinfo`
- `cacheinfo`
- `cacheutils`
- `termwriter`

Thoughts?

ashearin: Thats a good point. I was really trying to avoid a 'utils' folder and format was the first thing that came into my head. I think `outpututils` is the best of the bad ones. Still open to others though if anyone thinks of something better. 

As far as splitting table and list, the only function that mentions list is `formatList` but it's a receiver for the table struct so I dont think it warrants it's own file.

lmphil (approved):
Looks good

internal/pkg/outpututils/table.go
jhoward-lm, line 1: We don't have a convention documented, but I like ordering things in the file the same way they would show up in, for example, the VSCode outline.

In order:

- public methods
- private methods
- public functions
- private functions

with each category alphabetized. Also recommend renaming `getRow` --> `newRow` and `getTermInfo` --> `termInfo`. With that in mind, my suggestion is to use this order:

- `(t *Table) AddRow`
- `(t *Table) String`
- `(t *Table) determineColumnWidths`
- `(t *Table) formatList`
- `(t *Table) formatTable`
- `(t *Table) getHeaders`
- `(t *Table) getRows`
- `(t *Table) getTableWidth`
- `NewTable`
- `newRow`
- `termInfo`

jhoward-lm (approved):

PR #171, opened 2024-09-25
feat: testscript POC (DO NOT MERGE)

ashearin:
## Description

POC of testscript IT, for illustration/comparison purposes

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -172,6 +172,16 @@ func rootCmd() *cobra.Command {
 	return rootCmd
 }
 
-func Execute() {
-	cobra.CheckErr(rootCmd().Execute())
+// func Execute() {
+// 	cobra.CheckErr(rootCmd().Execute())
+// }
+
+// control your exit status with Execute, by default in cmd/root.go:.
+func Execute() int {
+	if err := rootCmd().Execute(); err != nil {
+		// log it, then
+		return 1
+	}
+
+	return 0
 }
--- a/internal/e2e/e2eutil/check_files.go
+++ b/internal/e2e/e2eutil/check_files.go
@@ -1,0 +1,49 @@
+// -----------------------------------------------------------------------------
+// SPDX-FileCopyrightText: Copyright © 2024 bomctl a Series of LF Projects, LLC
+// SPDX-FileName: internal/e2e/e2eutil/check_files.go
+// SPDX-FileType: SOURCE
+// SPDX-License-Identifier: Apache-2.0
+// -----------------------------------------------------------------------------
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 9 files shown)


internal/e2e/e2eutil/check_files.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"path"
+
+	"github.com/rogpeppe/go-internal/testscript"
+)
+
+// checkFile is a testscript command that checks the
+// existence of a list of files inside a directory.
+func checkFiles(script *testscript.TestScript, neg bool, args []string) { //nolint: revive
```
jhoward-lm, line 30: Does it have to match this signature? Or would this work. I also think I remember reading somewhere that `//nolint` directives shouldn't have a space after the `:`

```suggestion
func checkFiles(script *testscript.TestScript, neg bool, args ...string) { //nolint:revive
```

If it has to match this signature, and it's used in multiple places, can we define a type with it?

```go
type commandFunc func(*testscript.TestScript, bool, []string)
```

internal/e2e/e2eutil/custom_commands.go
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
+package e2eutil
+
+import "github.com/rogpeppe/go-internal/testscript"
+
+func CustomCommands() map[string]func(ts *testscript.TestScript, neg bool, args []string) {
+	return map[string]func(ts *testscript.TestScript, neg bool, args []string){
```
jhoward-lm, line 25: Same question as above about the signature

jhoward-lm, line 25: Shouldn't need to name the params if just used as a return type

```suggestion
func CustomCommands() map[string]func(*testscript.TestScript, bool, []string) {
	return map[string]func(*testscript.TestScript, bool, []string){
```

internal/e2e/e2eutil/custom_conditions.go
```diff
@@ -0,0 +1,130 @@
...
+		elapsed++
+	}
+
+	return false, errFileNotFound
+}
+
+func fileExists(filename string) bool {
+	_, err := os.Stat(filename)
+
+	return !os.IsNotExist(err)
```
jhoward-lm, line 129: From the [`os` docs](https://pkg.go.dev/os#IsNotExist)

```suggestion
	return !errors.Is(err, fs.ErrNotExist)
```

cmd/root.go
```diff
@@ -172,6 +172,16 @@ func rootCmd() *cobra.Command {
...
+// }
+
+// control your exit status with Execute, by default in cmd/root.go:.
+func Execute() int {
+	if err := rootCmd().Execute(); err != nil {
+		// log it, then
+		return 1
+	}
+
+	return 0
```
jhoward-lm, line 186: Doesn't main.go need updating?

There are several possible panics or `log.Fatal` calls in the subcommands. Should this also have a recover handler?

```suggestion
func Execute() (rc int) {
	defer func() {
		if r := recover(); r != nil {
			// handle the panic
			rc = 1
		}
	}()

	if err := rootCmd().Execute(); err != nil {
		rc = 1
	}

	return
```

internal/e2e/e2eutil/check_files.go
ashearin: you are right about the no lint thing, however we have two rules that argue over that, so I've been needing to put two nolint entries to get it to ignore the line

PR #202, opened 2024-11-05
fix: update compare_docs function

### LINKED ISSUE #178: bug: Fix spdx Metadata Equality issue seen in e2e tests
Export e2e tests uncovered a weird behavior in the testing environment, the spdx bom used in the export test is added to the cache via `db.AddDocument` and exported (to spdx format) to a file. the two files are then compared for equality, but fail on Meta data being different. 

Some things are expected, like `date` will always be different, which easily handled, but the ID, tools and author being different needs more investigation. 

```bash
Exported Document
# id:"https://spdx.org/spdxdocs/#DOCUMENT" version:"0" name:"bomctl_0.3.0_linux_amd64.tar.gz" date:{seconds:1727723887} tools:{name:"protobom-devel"} tools:{name:"syft-1.9.0"}
Original Document
# id:"https://anchore.com/syft/file/bomctl_0.3.0_linux_amd64.tar.gz-1b838d44-9d3c-47d0-9f7f-846397e701fa#DOCUMENT" version:"0" name:"bomctl_0.3.0_linux_amd64.tar.gz" date:{seconds:1723065476} tools:{name:"syft-1.9.0"} authors:{name:"Anchore, Inc" is_org:true}
```

The task here is to investigate and fix issue if it resides in bomctl, otherwise create an issue with supporting documentation and evidence in protobom to have it investigated there. 

Also update the `compare_documents` to compare each field individually, and ignore expected differences , like date. If resolved within bomctl, uncomment e2e test checks for document equality.

ashearin:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #178 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] Unit Tests
- [x] e2e tests

## Checklist


internal/e2e/e2eutil/compare_documents.go
```diff
@@ -99,18 +107,98 @@ func getFile(script *testscript.TestScript, filePath string) *os.File {
 	return file
 }
 
-func compareMetaData(script *testscript.TestScript, first, second *sbom.Metadata) bool {
-	firstStr := first.String()
-	secondStr := second.String()
+func compareMetaData(script *testscript.TestScript, have, want *sbom.Metadata) bool {
+	equal := true
+
+	if have.GetId() != want.GetId() {
```
jhoward-lm, line 113: Probably want to use a switch case here, so it doesn't need to check every single condition. The first time it hits false, it breaks out

```go
switch {
case have.GetId() != want.GetId():
    // log and return `false`
...
}
```

internal/e2e/e2eutil/compare_documents.go
```diff
@@ -99,18 +107,98 @@ func getFile(script *testscript.TestScript, filePath string) *os.File {
 	return file
 }
 
-func compareMetaData(script *testscript.TestScript, first, second *sbom.Metadata) bool {
-	firstStr := first.String()
-	secondStr := second.String()
+func compareMetaData(script *testscript.TestScript, have, want *sbom.Metadata) bool {
```
jhoward-lm, line 110: Could this not just be something like

```go
haveBytes, err := proto.MarshalOptions{Deterministic: true}.Marshal(have)
if err != nil {
    // handle error
}

wantBytes, err := proto.MarshalOptions{Deterministic: true}.Marshal(want)
if err != nil {
    // handle error
}

return haveBytes == wantBytes
```

ashearin: Quick test didn't work, will dig a bit more. Would this handle the timestamp? Thats the biggest reason for doing the individual field checks

ashearin: I think for the time being we'll have to keep the individual checks, the source data hash and the timestamp are going to always be different

jhoward-lm: In the past for situations like this, I've removed the timestamp from both before comparison. It doesn't look like you're comparing timestamps between the two anyway, so it's effectively the same thing

internal/e2e/e2eutil/compare_documents.go
jhoward-lm: If implementing the solution from my other comment, this comment should be ignored

internal/e2e/e2eutil/compare_documents.go
ashearin: Oh cool, didn't think of doing that. I'll make that change

ashearin: Ok so there are 3 fields that are expected to be different. (apologies for the back and forth, it's been a bit since I looked at this originally)

1. MetaData.SourceData.Hash (and presumably size)
2. MetaData.Date
3. MetaData.Tools (protobom appends an entry here)

Following your example, I was able to get it working, with a couple additional steps to account for all 3 fields listed above. I'll push that change momentarily. It should be noted that the addition of a protobom tool will not be needed once the without modification changes take place so i wonder if it's better to nil out that field as well. Thoughts?

internal/e2e/e2eutil/compare_documents.go
```diff
@@ -99,18 +110,84 @@ func getFile(script *testscript.TestScript, filePath string) *os.File {
...
+	have.SourceData = nil
+	want.SourceData = nil
+
+	// remove timestamp to ignore time created mismatch
+	have.Date = nil
+	want.Date = nil
+
+	haveBytes, err := proto.MarshalOptions{Deterministic: true}.Marshal(have)
+	if err != nil {
+		script.Fatalf("failed to Marshal MetaData: %v", err)
```
jhoward-lm, line 131: nit: casing

```suggestion
		script.Fatalf("failed to marshal metadata: %v", err)
```

internal/e2e/e2eutil/compare_documents.go
```diff
@@ -99,18 +110,84 @@ func getFile(script *testscript.TestScript, filePath string) *os.File {
...
+	}
 
-	script.Logf(firstStr)
-	script.Logf(secondStr)
+	wantBytes, err := proto.MarshalOptions{Deterministic: true}.Marshal(want)
+	if err != nil {
+		script.Fatalf("failed to Marshal MetaData: %v", err)
```
jhoward-lm, line 136: nit: casing

```suggestion
		script.Fatalf("failed to marshal metadata: %v", err)
```

internal/e2e/e2eutil/compare_documents.go
```diff
@@ -99,18 +110,84 @@ func getFile(script *testscript.TestScript, filePath string) *os.File {
...
+		want.Tools = append(want.GetTools(), &sbom.Tool{Name: "protobom-devel"})
+		slices.SortFunc(want.GetTools(), func(i, j *sbom.Tool) int {
+			return strings.Compare(i.GetName(), j.GetName())
+		})
+	}
+	// remove sourceData to avoid hash mismatches due to format specific losses
+	have.SourceData = nil
+	want.SourceData = nil
+
+	// remove timestamp to ignore time created mismatch
```
jhoward-lm, line 125: nits: casing, end comments with period, blank line after `if` block

```suggestion

	// Remove SourceData to avoid hash mismatches due to format specific losses.
	have.SourceData = nil
	want.SourceData = nil

	// Remove Date to ignore time created mismatch.
```

internal/e2e/e2eutil/compare_documents.go
```diff
@@ -64,26 +73,28 @@ func compareDocuments(script *testscript.TestScript, neg bool, args []string) {
...
-		}
-
 		script.Fatalf("documents do not match")
 	}
 
 	if docsMatch && neg {
 		script.Fatalf("documents Match")
 	}
+
+	script.Logf("documents Match")
```
jhoward-lm, line 97: nit: casing

```suggestion
		script.Fatalf("documents match")
	}

	script.Logf("documents match")
```

jhoward-lm (approved):

internal/e2e/e2eutil/compare_documents.go
jhoward-lm: LGTM!

ashearin: Also as an side comment, the addition of `protobom-devel` Tool only happens in spdx documents, as my code suggests, but I'm not sure why that is. maybe worth an issue in protobom?

And I think it's supposed to be `protobom-v0.5.0` (or match the go mod in general) and I've only ever seen `protobom-devel`. Could be an easy bug to fix also

PR #210, opened 2024-11-13
feat: Output Sboms without Modification

### LINKED ISSUE #187: Storing and Retrieving Source SBOMs Without Modification
[Link](https://github.com/bomctl/bomctl/blob/main/ROADMAP.md#storing-and-retreiving-source-sboms-without-modification-2-months) to Roadmap.md with more details. 

TLDR: There are situations where the original imported/fecthed document should be push/exported without changes. 

We currently save off the original document (as a string) and document format as unique annotations in the DB. It should only be a matter of checking these fields upon export/push to see if the desired format is the same as the original format and if it is use the original document string in lieu of protobom document. 

SInce this may be mainly a signature preservation issue, maybe a flag like `preserve-signatures` would push it as the origin document without additional flags. Would also work in situations where the original format is not readily known.

ashearin:
## Description

Adds infrastructure and special format flag to output sboms as original source content if appropriate. More information can be found in issue linked below.

** Dependent on a release to protobom storage ** 

Fixes #187 

## Type of change

## Checklist


ashearin:
Dependent on https://github.com/protobom/storage/pull/42

ashearin:
> Dependent on [protobom/storage#42](https://github.com/protobom/storage/pull/42)

Associated storage PR has been merged, updated this branch to point at the master branch for now until a new release is cut so review can be completed

ashearin:
I think that makes sense, It'll output the source bytes unless changes are made. Also aligns with 'format agnostic'

internal/pkg/outpututil/file.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+
+	modified, err := checkIfModified(document, backend)
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+
+	if string(format) == db.OriginalFormat || (!modified && matchesOriginFormat(document, format)) {
```
jhoward-lm, line 72: If the suggestion in db.go works, this would probably then just be

```suggestion
	if format == db.OriginalFormat || (!modified && matchesOriginFormat(document, format)) {
```

internal/pkg/db/db.go
```diff
@@ -49,6 +49,8 @@ const (
 	DatabaseFile string = "bomctl.db"
 
 	EntDebugLevel int = 2
+
+	OriginalFormat = "original"
```
jhoward-lm, line 53: In the call, what I was suggesting was

```suggestion
	OriginalFormat = formats.Format("original")
```

Will that work? Or does it still require casting to string

jhoward-lm (approved):

internal/pkg/db/db.go
ashearin: It works the same as the initial version. meaning it requires casting to a string in `cmd/export.go`. I went this way because the `formats.SPDX` and the like are untyped strings in protobom, so this matches that pattern

internal/pkg/outpututil/file.go
ashearin: So the linter didn't even complain about this if I didn't cast it to a string, and the tests passed. so this might be optional, but it seemed weird to me that it didn't complain

lallevato-lm (approved):

internal/pkg/outpututil/file.go
jhoward-lm: I think because `formats.Format` has an underlying type of `string` (i.e. because it's defined as `type Format string`), Go lets you compare strings against it directly

ashearin: I appreciate the quick exmaple! I'll make this change and we should be gtg

internal/pkg/db/db.go
ashearin: I'm going to keep it as an untyped string and remvoe the cast in your other comment. I think thats the best option

Issue #81, opened 2024-06-18
SBOM "link" feature


nishakm:
## Summary
Given one or more SBOM documents, create a third SBOM document containing external references to those documents.
Something like:
```
bomctl link sbom1.json sbom2.json ... -o linked_sbom.json
```
Ideally, bomctl should also be able to express relationships between the components in the sbom:
```
bomctl link relationship=from:sbom1.json,to:sbom2.json,type:contains
```
## Use Case
Suppose a base container image has an SBOM
```mermaid
graph LR
    A(Distroless Creators)
    B(Distroless container image)
    C(Distroless SBOM)
    A --> B
    A --> C
```
And a Go binary has its own SBOM
```mermaid
graph LR
    A(Go Binary Creators)
    B(Go Binary)
    C(Go binary SBOM)
    A --> B
    A --> C
```
During containerization of the Go binary, I would also like to create an SBOM for this container which has external references to the base container's SBOM and the go binary's SBOM
```mermaid
graph LR
    A(Distroless Container)
    B(Go Binary)
    C(Containerized Go Binary)
    A -->|FROM distroless:latest| C
    B --> |COPY go_binary|C

    D(Containerized Go Binary SBOM)
    E(Distroless Container SBOM)
    F(Go Binary SBOM)

    D --> |contains|E
    D --> |contains|F
    F --> |dependsOn|E
```

PR #211, opened 2024-11-14
feat: add base document db annotation, WithSourceDocumentAnnotations and WithRevisedDocumentAnnotations functions

### LINKED ISSUE #203: Add / Define base document database UUID annotation
Need to define annotation that defines the base document for modified or changed documents.

baseDocumentUUID -> prior document's database id

The annotation should remain off any fetched / imported documents and added to any new documents that are generated from / based on those fetched / imported documents.

ashearin:
## Description

Adds an annotation for storing the uuid of the 'base' document when creating a new, modified document from one imported/fetched into the cache. In the added function the new document is called a revision. 

All names are open to suggestions if you feel they don't capture the intent properly.

Also added a AddDocumentRevision function that takes in two sbom documents, the base document and the newly created revision document and setting appropriate annotations, including the base document uuid annotation, and moves the alias 'pointer' to the latest revision of the document, if an alias is present on the original. 

Fixes #203
Related to #81 

## Type of change

## How Has This Been Tested?

- [x] Expanded Unit tests

## Checklist


internal/pkg/db/db.go
```diff
@@ -238,6 +270,30 @@ func (backend *Backend) FilterDocumentsByTag(documents []*sbom.Document, tags ..
...
 
+func (backend *Backend) MoveAlias(revision, base *sbom.Document) error {
+	baseUUID := base.GetMetadata().GetId()
+
+	docAlias, err := backend.GetDocumentUniqueAnnotation(baseUUID, AliasAnnotation)
+	if err != nil {
+		return fmt.Errorf("failed checking for existing alias: %w", err)
+	}
+
+	// if parent has existing alias, move to child
```
jhoward-lm, line 281:
```suggestion
	// If parent has existing alias, move to child.
```

internal/pkg/db/db.go
```diff
@@ -238,6 +270,30 @@ func (backend *Backend) FilterDocumentsByTag(documents []*sbom.Document, tags ..
...
+	if docAlias != "" {
+		// remove alias from parent document
```
jhoward-lm, line 283:
```suggestion
		// Remove alias from parent document.
```

internal/pkg/db/db.go
```diff
@@ -238,6 +270,30 @@ func (backend *Backend) FilterDocumentsByTag(documents []*sbom.Document, tags ..
...
+		if err := backend.RemoveAnnotations(baseUUID, AliasAnnotation, docAlias); err != nil {
+			return fmt.Errorf("failed to remove existing alias: %w", err)
+		}
+
+		// set alias on child document
```
jhoward-lm, line 288:
```suggestion
		// Set alias on child document.
```

internal/pkg/db/db.go
```diff
@@ -238,6 +270,30 @@ func (backend *Backend) FilterDocumentsByTag(documents []*sbom.Document, tags ..
 	return documents, nil
 }
 
+func (backend *Backend) MoveAlias(revision, base *sbom.Document) error {
+	baseUUID := base.GetMetadata().GetId()
```
jhoward-lm, line 274: This isn't a UUID, it's the metadata ID string. Currently it's unique since it's the primary key but I think the eventual goal is to change that. Might need another update to the storage backend to allow for operations on UUIDs

If a UUID is what you need you should be able to call

```go
baseUUID, err := ent.GenerateUUID(base)
```

internal/pkg/db/db.go
```diff
@@ -238,6 +270,30 @@ func (backend *Backend) FilterDocumentsByTag(documents []*sbom.Document, tags ..
 	return documents, nil
 }
 
+func (backend *Backend) MoveAlias(revision, base *sbom.Document) error {
```
jhoward-lm, line 273: Something like the name `UpdateAlias` might make more sense, how would you feel about that?

internal/pkg/db/db_test.go
```diff
@@ -56,6 +56,75 @@ func (dbs *dbSuite) TearDownSuite() {
 	dbs.Backend.CloseClient()
 }
 
+func (dbs *dbSuite) TestAddDocumentRevision() {
```
jhoward-lm, line 59:
```suggestion
func (dbs *dbSuite) TestBackend_AddDocumentRevision() {
```

internal/pkg/db/db_test.go
```diff
@@ -56,6 +56,75 @@ func (dbs *dbSuite) TearDownSuite() {
...
+	baseUUID := dbs.documents[1].GetMetadata().GetId()
+	revisionUUID := dbs.documents[0].GetMetadata().GetId()
+
+	tempbe, err := testutil.NewTestBackend()
+	dbs.Require().NoError(err, "failed database backend creation")
+	defer tempbe.CloseClient()
```
jhoward-lm, line 65: Might be good candidates for `SetupSubTest` and `TearDownSubTest`

internal/pkg/db/db_test.go
```diff
@@ -154,8 +223,87 @@ func (dbs *dbSuite) TestFilterDocumentsByTag() {
 	}
 }
 
+func (dbs *dbSuite) TestMoveAlias() {
```
jhoward-lm, line 226:
```suggestion
func (dbs *dbSuite) TestBackend_MoveAlias() {
```

internal/pkg/db/db.go
ashearin: Yeah I also thought about TransferAlias, since thats the goal of the function

internal/pkg/db/db_test.go
ashearin: I'm going to look more into this. The tests in this file specifically expect a pristine db, in a state after SetupSuite(), which is how it should be, but me monkeying around with annotations and aliases to test them is impacting others, so I'll figure out a way to keep everything tidy

ashearin: Ending up using Setup/Teardown Subtest functions, but I moved my test additions to a different test file, since they are the only tests making annotation/alias adjustments. lmk what you think

internal/pkg/db/db.go
```diff
@@ -102,8 +103,39 @@ func NewBackend(opts ...Option) (*Backend, error) {
...
+					IsUnique: true,
+				},
+				{
+					Name:     SourceHashAnnotation,
+					Value:    string(hash[:]),
+					IsUnique: true,
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
ashearin: Is the UUID stored anywhere in the sbom document, or is it just known to the db? Regardless, I guess I have two questions about your suggestion:
- Would we not be generating that UUID for the base document upon adding it (the base doc) to the db upon fetch or import? (So in AddOriginDocument)
- If I run GetDocumentByID and pass it the base UUID, would that return the correct doc?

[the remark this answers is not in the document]
jhoward-lm: > Is the UUID stored anywhere in the sbom document, or is it just known to the db?

It's database-specific and doesn't exist as part of the protobom spec

> Would we not be generating that UUID for the base document upon adding it (the base doc) to the db upon fetch or import? (So in AddOriginDocument)

It's generated by the ent backend during `Store` but should be able to be deterministically recreated from bomctl by calling `ent.GenerateUUID()`

> If I run GetDocumentByID and pass it the base UUID, would that return the correct doc?

No, they are two different types, `GetDocumentByID` expects a string. There are still changes in protobom/storage that will need to be made to make it official, but the string ID will no longer represent a distinct document. Probably also need to add a new method in protobom/storage to get a document by UUID for that use case

jhoward-lm:
From a design perspective, to reduce duplication, how would you feel about this?

- Keep the original method `AddDocument` but update it to accept 0 or more annotations
- Update `AddOriginDocument` so that all it does is create a `[]*ent.Annotation` (or `ent.Annotations`, same thing) and call `AddDocument` with it. No need to call `Store` from multiple places
- Rename `AddOriginDocument` --> `AddSourceDocument` to stick with the alias name conventions (e.g. `bomctl_annotation_source_format`, `bomctl_annotation_source_hash`, etc.)
- Replace `AddDocumentRevision` with a method like `CreateBaseDocumentAnnotation` that just generates the UUID for a document and returns the annotation that can be passed to `AddDocument`

```go
func (backend *Backend) AddDocument(sbomData []byte, ...annotations *ent.Annotation) (*sbom.Document, error) {
    // create ent.BackendOptions with the annotations slice

    // call backend.Store
}

func (backend *Backend) AddSourceDocument(sbomData []byte) (*sbom.Document, error) {
    // create slice of format, hash, and data annotations

    // call backend.AddDocument(sbomData, annotations...)
}

func (backend *Backend) CreateBaseDocumentAnnotation(document *sbom.Document) (*ent.Annotation, error) {
    baseUUID, err := ent.GenerateUUID(base)
    if err != nil {
        return nil, fmt.Errorf("%w", err)
    }

    return &ent.Annotation{
        Name: BaseDocumentAnnotation,
        Value: string(baseUUID),
        IsUnique: true,
    }
}
```

internal/pkg/db/db.go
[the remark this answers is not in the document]
jhoward-lm: I want to pump the brakes a little so we don't lock ourselves into any design patterns (stuff like [this comment](https://github.com/bomctl/bomctl/pull/211#issuecomment-2481424762)). I don't want to merge as-is until I have a full picture of what all needs to be done upstream, and I don't see this as super urgent or anything

internal/pkg/db/db.go
jhoward-lm: I think I want to revisit this. To me, `TransferAlias` doesn't seem to really describe its purpose. When I think "transfer", I think copy data from one place to another. This is more like updating the underlying reference for a pointer. So in my opinion, `UpdateAliasReference` makes the most sense, maybe abbreviated to just `UpdateAliasRef` or (less preferable) `UpdateAlias`

ashearin:
> From a design perspective, to reduce duplication, how would you feel about this?

I like this. I tried to do something with Interfaces (any) initially, but I felt like it made it messy and harder to follow so I went with separate functions. I'll make the switch to follow your example

ashearin:
> * Instead of creating `AddDocumentRevision`, I'd prefer a method like `CreateBaseDocumentAnnotation` that just generates the UUID for a document and returns the annotation that can then be passed to `AddDocument`

Running through the logistics of your suggestion. Were you thinking the creation of a 'revised' document would look something like:
```go
revision := sbom.NewDocument() // or whatever

baseDocAntn, err := be.CreateBaseDocumentAnnotation(base)
// check err

_, err = be.AddDocument(revision, baseDocAntn)
// check err

err = be.UpdateAliasReference(revision, base)
// check err
```

One of reasons I thought `AddDocumentRevision` was worth creating was to only have one call to a function and one error to check like: 
```go
revision := sbom.NewDocument() // or whatever

 _, err := be. AddDocumentRevision(revision, base)
// check err
```
And internal to `AddDocumentRevision` would be those three function calls it replaced. We just call AddDocument inside `AddDocumentRevision` with the appropriate annotations like we do in`AddSourceDocument`, keeping only the single call to store. 

Also reading ahead (AKA I already have this code written, minus the updates from this conversation), when we add a 'latest' document annotation we'll have another function we have to call (to move the latest annotation to the revised doc) when we create a revised document that could be contained in `AddDocumentRevision`, keeping things clean and avoiding repeated calls. 

Not sure how many places we'll be eventually creating document revisions, but I'm assuming it'd be in more than one place. 

Could also switch it up to `AddRevisionDocument` to align the function naming

lmphil (approved):

lallevato-lm (approved):

jhoward-lm (approved):

jhoward-lm:
Documenting discussion of this PR in an earlier call:

As a future effort, it might make sense to implement a functional options pattern for passing in a set of annotations to `Backend.AddDocument`

```go
func WithSourceDataAnnotation(sbomData []byte) Option {
	return func(backend *Backend) {
		backend.Annotations = append(backend.Annotations, ent.Annotations{
			{
				Name:     SourceDataAnnotation,
				Value:    string(sbomData),
				IsUnique: true,
			},
		})
	}
}
```

or a combination of predefined annotations for different scenarios

internal/pkg/db/db.go
```diff
@@ -104,31 +109,31 @@ func NewBackend(opts ...Option) (*Backend, error) {
 	return backend, nil
 }
 
-// AddDocument adds the protobom Document to the database and annotates with its source data, hash, and format.
-func (backend *Backend) AddDocument(sbomData []byte) (*sbom.Document, error) {
+// AddDocument adds the protobom Document to the database and annotates with given annotations.
```
ashearin, line 112:
```suggestion
// AddDocument adds the protobom document to the database and annotates with given annotations.
```

internal/pkg/db/db.go
```diff
@@ -279,16 +284,88 @@ func (backend *Backend) validateNewAlias(alias string) (err error) {
...
+				IsUnique:   true,
+			},
+		)
+
+		docAlias, err := backend.GetDocumentUniqueAnnotation(baseID, AliasAnnotation)
+		if err != nil {
+			return fmt.Errorf("failed checking for existing alias: %w", err)
+		}
+
+		// If parent has existing alias, move to child.
```
ashearin, line 333:
```suggestion
		// If base doc has existing alias, move to revised doc.
```

internal/pkg/db/db.go
```diff
@@ -279,16 +284,88 @@ func (backend *Backend) validateNewAlias(alias string) (err error) {
...
+		if docAlias != "" {
+			// Remove alias from parent document.
+			if err := backend.RemoveDocumentAnnotations(baseID, AliasAnnotation, docAlias); err != nil {
+				return fmt.Errorf("failed to remove existing alias: %w", err)
+			}
+
+			// Add Alias Annotation to new document
```
ashearin, line 340:
```suggestion
			// Add AliasAnnotation to revised document
```

internal/pkg/db/db.go
```diff
@@ -279,16 +284,88 @@ func (backend *Backend) validateNewAlias(alias string) (err error) {
...
+		)
+
+		docAlias, err := backend.GetDocumentUniqueAnnotation(baseID, AliasAnnotation)
+		if err != nil {
+			return fmt.Errorf("failed checking for existing alias: %w", err)
+		}
+
+		// If parent has existing alias, move to child.
+		if docAlias != "" {
+			// Remove alias from parent document.
```
ashearin, line 335:
```suggestion
			// Remove alias from base document.
```

internal/pkg/db/db.go
```diff
@@ -104,31 +109,31 @@ func NewBackend(opts ...Option) (*Backend, error) {
...
 	if err != nil {
 		return nil, fmt.Errorf("parsing SBOM data: %w", err)
 	}
 
-	hash := sha256.Sum256(sbomData)
+	// Loop over the `annotations` slice and call each one
+	for _, fn := range annotations {
+		err := fn(backend)
+		if err != nil {
+			return nil, fmt.Errorf("handling annotations: %w", err)
```
ashearin, line 130:
```suggestion
			return nil, fmt.Errorf("handling document annotations: %w", err)
```

internal/pkg/db/db.go
```diff
@@ -104,31 +109,31 @@ func NewBackend(opts ...Option) (*Backend, error) {
...
+		}
+	}
+
+	// Create ent.BackendOptions with the annotations slice.
```
ashearin, line 134:
```suggestion
	// Create StoreOptions with the populated backend options struct.
```

lallevato-lm (approved):

internal/pkg/db/db.go
```diff
@@ -104,31 +109,31 @@ func NewBackend(opts ...Option) (*Backend, error) {
 	return backend, nil
 }
 
-// AddDocument adds the protobom Document to the database and annotates with its source data, hash, and format.
-func (backend *Backend) AddDocument(sbomData []byte) (*sbom.Document, error) {
+// AddDocument adds the protobom document to the database and annotates with given annotations.
+func (backend *Backend) AddDocument(sbomData []byte, annotations ...Option) (*sbom.Document, error) {
+	// defer clearing annotations for next run
```
jhoward-lm, line 114: nits:
- casing
- If the code has `defer` in it, the comment shouldn't need to restate that
- the wording could be a little more focused on "why is it doing this" rather than "what is it doing"
  - e.g. `// Clear annotations after adding document.` 

```suggestion
	// Defer clearing annotations for next run.
```

internal/pkg/db/db.go
```diff
@@ -104,31 +109,31 @@ func NewBackend(opts ...Option) (*Backend, error) {
...
+
 	sbomReader := reader.New()
 
 	document, err := sbomReader.ParseStream(bytes.NewReader(sbomData))
 	if err != nil {
 		return nil, fmt.Errorf("parsing SBOM data: %w", err)
 	}
 
-	hash := sha256.Sum256(sbomData)
+	// Loop over the `annotations` slice and call each one
```
jhoward-lm, line 126: Again, "why is it doing this" rather than "what is it doing"

internal/pkg/db/db_revised_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+				alias, err := dbrs.Backend.GetDocumentUniqueAnnotation(newID, db.AliasAnnotation)
+				dbrs.Require().NoError(err)
+				dbrs.Require().Equal(data.alias, alias)
+
+				alias, err = dbrs.Backend.GetDocumentUniqueAnnotation(baseDocID, db.AliasAnnotation)
+				dbrs.Require().NoError(err)
+				dbrs.Require().Equal("", alias)
+
+				soureData, err := dbrs.Backend.GetDocumentUniqueAnnotation(newID, db.SourceDataAnnotation)
```
jhoward-lm, line 108: `sourceData` (missing "c") or just `srcData`

```suggestion
				sourceData, err := dbrs.Backend.GetDocumentUniqueAnnotation(newID, db.SourceDataAnnotation)
```

internal/pkg/db/db_revised_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/bomctl/bomctl/internal/pkg/db"
+	"github.com/bomctl/bomctl/internal/testutil"
+)
+
+type dbrSuite struct {
+	suite.Suite
+	*db.Backend
+	documentInfo []testutil.DocumentInfo
+}
```
jhoward-lm, line 35: Something about `db_revised_test.go` and `dbrSuite` is nagging at me, hard to pin down. I think the word `revised` is throwing me off, like what is revised referring to, the database? (I know it's referring to a revised document, just the semantics are a little unclear)

To me, it makes more sense to move all of this into `db_test.go`. There are a few things you can try to compact the code and minimize redundancy here and there, such as embedding `dbSuite` into `dbrSuite`:

```go
type (
	dbSuite struct {
		suite.Suite
		*db.Backend
		documents    []*sbom.Document
		documentInfo []testutil.DocumentInfo
	}

	dbRevSuite struct {
		dbSuite
	}

	dbStdSuite struct {
		dbSuite
	}
)
```

`dbSuite` isn't actually run with `suite.Run(t, new(dbSuite))`, it's just to define the common fields and any shared setup/teardown methods you want to implement (`BeforeTest`, `SetupTestSuite`, etc.)

You can probably call both sub-suites from one test function (or separate test functions if you prefer)

```go
func TestDBSuite(t *testing.T) {
	t.Parallel()
	suite.Run(t, new(dbStdSuite))
	suite.Run(t, new(dbRevSuite))
}
```

What are your thoughts?

jhoward-lm: Be careful with embedding though, I just realized I made a mistake embedding `gitClientSuite` into `gitPushSuite` and `gitFetchSuite`, it's running the parse tests defined on the client suite as if all 3 suites had defined them independently 🤦

internal/pkg/db/db.go
jhoward-lm: Also the variable should probably be renamed to `opts` or `backendOpts` or something

internal/pkg/db/db_revised_test.go
ashearin: Consolidated them into db_test.go, lmk what you think

internal/pkg/db/db.go
ashearin: Resolved the name, updated the comment. Let me know what you think

jhoward-lm:
I believe for [this commit](https://github.com/bomctl/bomctl/pull/211/commits/f960596d1529951f2a2a0d0c6b9267f87a2a37cf), leaving it unset will cause the protobom/storage schema to randomly generate one, which won't match any document IDs in the database

internal/pkg/db/db_revised_test.go
jhoward-lm: Looks great!

ashearin:
> I believe for [this commit](https://github.com/bomctl/bomctl/pull/211/commits/f960596d1529951f2a2a0d0c6b9267f87a2a37cf), leaving it unset will cause the protobom/storage schema to randomly generate one, which won't match any document IDs in the database

What would I set it to in this situation? Would they all be the new/revised document uuid?

ashearin:
> Should be `uuid.Nil`. From protobom/storage `Store` method
> 
> ```go
> 	// Set each annotation's document ID if not specified.
> 	for _, a := range annotations {
> 		if a.DocumentID == uuid.Nil {
> 			a.DocumentID = id
> 		}
> 	}
> ```

Ok cool, was just looking at that. Thanks for the verification. reverted that commit. Should be gtg

PR #219, opened 2024-11-19
docs: ADR-0009:  url refactor

ashearin:
## Description

ADR documentation for updating and refinement of URL parsing and URL/Location syntax

Added it as a folder for now with additional files to make it easier to review, can consolidate before merging,

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/docs/architecture/009-use-standard-for-url-structure/0009-supporting-packageurl.md
+++ b/docs/architecture/009-use-standard-for-url-structure/0009-supporting-packageurl.md
@@ -98,7 +98,7 @@ pkg:rpm/opensuse/curl@7.56.1-1.1.?arch=i386&distro=opensuse-tumbleweed
 - Git  (Although all these examples could also have the prefix `pkg:github` and use the future github client)
   - `pkg:git/bomct/bomctl@main?user=git#sbom.cdx.json`
   - `pkg:git/bomct/bomctl@main?user=git&port=12345&scheme=ssh#sbom.cdx.json`
-  - `pkg:git/bomct/bomctl@main?user=git&password=password&port=12345&scheme=git#sbom.cdx.json`
+  - `pkg:git/bomct/bomctl@main?user=git&password=password&port=12345#sbom.cdx.json`
   - `pkg:git/bomct/bomctl@main?user=git#sbom.cdx.json`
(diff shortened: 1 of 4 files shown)


jhoward-lm:
This looks good! I think it should be all in one document conforming to the ADR spec, even if it is large. PR description should be the content of that document.

You can use this markdown pattern to collapse sections

```markdown
<details>
<summary>Example</summary>

<!-- Collapsed content -->

</details>
```

ashearin:
> You can use this markdown pattern to collapse sections

Good Idea! I'll switch it up

jhoward-lm:
Update casing of acronyms throughout, e.g. `HTTP`, `URLs`, `SBOMs` (I think `purl` is ok because that's how it shows up in the purl-spec documentation)

docs/architecture/009-use-standard-for-url-structure/0009-supporting-packageurl.md
```diff
@@ -0,0 +1,106 @@
+# Investigation of using PackageURL as Command 'url'/location standard
+
+## Description
```
jhoward-lm, line 3: I think for this whole section it might be ok to just link to the purl-spec documentation

lmphil (approved):

lallevato-lm (commented):
Had a thought on this ADR. Regardless of what solution is implemented here, it is probably worth looking into adding Cobra's [autocomplete](https://github.com/bomctl/bomctl/pull/238) when typing in the URLs for a nice QoL enhancement. For example, in the current parsing scheme, it could show you the schemes that could be completed with your current input. So if you typed 

`bomctl push SBOM_ID git://username:password [tab][tab]` 

Then perhaps autocomplete would spit out 

`(scheme)://(username):(password)@(hostname):(port)/(path)`

Or something similar, as well as any other URL schemes which match the pattern you have already typed. This way the user doesn't have to be looking at the documentation for how to format the URL.

ashearin:
Discussed in todays working meeting: 
- PURL implementation likely to cause users to manually generate said PURLs, vs a copy paste of an asset location
- Need to balance magic/inferring the users intention and requiring explicit client specification
- Do Skopeo/Trivy/Oras use a known standard? is it with aligning with their format? Does this only apply to OCI client?

PR #229, opened 2024-11-27
feat: gitlab fetch

### LINKED ISSUE #131: Initial Gitlab Client Skeleton and Sbom api Fetch
Gitlab api could be leveraged to perform bomctls i/o subcommands like fetch and push. Our users will mainly be using github/gitjab and should be given proper support.

There also ay be future api improvements that allow for importing sboms from gitlab projects.

lmphil:
## Description

This pull request adds the ability to fetch an SBOM from GitLab using the [Dependency List Export](https://docs.gitlab.com/ee/api/dependency_list_export.html) web API endpoint.

Fixes #131

## Type of change

## How Has This Been Tested?

- [x] Manual tests.
- [x] Unit tests.

## Checklist


internal/pkg/client/gitlab/fetch_test.go
jhoward-lm, line 1: Missing SPDX file header

internal/pkg/client/gitlab/fetch.go
```diff
@@ -0,0 +1,227 @@
...
+
+type dependencyListExportSession struct {
+	Client ClientWrapperInterface
+	Export *gitlab.DependencyListExport
+}
+
+var (
+	errInvalidGitLabURL = errors.New("invalid URL for GitLab fetching")
+	errFailedWebRequest = errors.New("web request failed")
+	errForbiddenAccess  = errors.New("you don't have permission to read the dependency list")
```
jhoward-lm, line 72: Avoid "you" statements

```suggestion
	errForbiddenAccess  = errors.New("the supplied token is missing the read_dependency permission")
```

internal/pkg/client/gitlab/fetch_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+var successGitLabResponse = &gogitlab.Response{
+	Response: &http.Response{
+		Status:     "200 OK",
+		StatusCode: 200,
+		Proto:      "HTTP/1.0",
+	},
+}
+
+func TestCreateExport(t *testing.T) { //nolint:paralleltest
```
jhoward-lm, line 83: These should be using a testify `suite.Suite` receiver. There's a VSCode snippet you can use to generate all the stubs for a suite, type `stretchr` and select the snippet suggestion to generate

internal/pkg/client/gitlab/push.go
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
+package gitlab
+
+import (
+	"github.com/bomctl/bomctl/internal/pkg/options"
+)
```
jhoward-lm, line 24:
```suggestion
import "github.com/bomctl/bomctl/internal/pkg/options"
```

internal/pkg/client/gitlab/fetch_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	if err := session.createExport("DUMMY_PROJECT", "DUMMY_BRANCH"); err != nil {
+		t.Errorf("failed to create dependency list export: %v", err)
+	}
+
+	mockedGoGitLabClient.AssertExpectations(t)
+}
+
+func TestPollExportUntilFinished(t *testing.T) { //nolint:paralleltest
```
jhoward-lm, line 166: Please avoid ignoring linters. Use a testify suite as suggested above and it won't be needed (need to fix multiple instances)

internal/pkg/client/gitlab/fetch_test.go
```diff
@@ -0,0 +1,201 @@
...
+	return args.Get(0).(*gogitlab.DependencyListExport), args.Get(1).(*gogitlab.Response), args.Error(2) //nolint:revive,wrapcheck,lll
+}
+
+func (mc *mockClient) DownloadDependencyListExport(
+	id int,
+	options ...gogitlab.RequestOptionFunc,
+) (io.Reader, *gogitlab.Response, error) {
+	args := mc.Called(id, options)
+
+	return args.Get(0).(io.Reader), args.Get(1).(*gogitlab.Response), args.Error(2) //nolint:revive,wrapcheck
```
jhoward-lm, line 72: For the `revive` linter, use a `//revive:disable` style directive instead to suppress only a specific rule, e.g. `//revive:disable:defer` (need to fix multiple instances)

jhoward-lm (dismissed):

jhoward-lm:
I didn't explain what I was getting at for implementing a mockable pattern very well during our call, so hopefully this will make more sense. In my opinion, this type of implementation might make the code easier to read, maintain, and test/mock.

The TL;DR version is even with nested types and interfaces, it might simplify things to abstract everything into a common flattened interface. This interface can then be implemented as a mock much more easily.

1. Define interfaces for methods you want to mock and embed these interfaces as fields on your `Client` struct definition

    <details>

    <summary>Example</summary>

    ```go
    // internal/pkg/client/gitlab/client.go

    type (
    	BranchesService interface {
    		GetBranch(pid any, branch string, options ...gitlab.RequestOptionFunc) (*gitlab.Branch, *gitlab.Response, error)
    	}
    
    	ProjectsService interface {
    		GetProject(pid any, opt *gitlab.GetProjectOptions, options ...gitlab.RequestOptionFunc) (*gitlab.Project, *gitlab.Response, error)
    	}
    
    	CommitsService interface {
    		GetCommit(pid any, sha string, opt *gitlab.GetCommitOptions, options ...gitlab.RequestOptionFunc) (*gitlab.Commit, *gitlab.Response, error)
    	}
    
    	DependencyListExportService interface {
    		CreateDependencyListExport(pipelineID int, opt *gitlab.CreateDependencyListExportOptions, options ...gitlab.RequestOptionFunc) (*gitlab.DependencyListExport, *gitlab.Response, error)
    		DownloadDependencyListExport(id int, options ...gitlab.RequestOptionFunc) (io.Reader, *gitlab.Response, error)
    		GetDependencyListExport(id int, options ...gitlab.RequestOptionFunc) (*gitlab.DependencyListExport, *gitlab.Response, error)
    	}

    	Client struct {
    		BranchesService
    		ProjectsService
    		CommitsService
    		DependencyListExportService
    	
    		glClient *gitlab.Client
    	}
    )
    ```

    </details>

2. If one of these interfaces is unset in the code before trying to call its interface method, set it to a default value

    <details>

    <summary>Example</summary>

    ```go
    func (client *Client) createExport(projectName, branchName string) error {
    	if client.ProjectsService == nil {
    		client.ProjectsService = client.glClient.Projects
    	}
    
    	project, response, err := client.ProjectsService.GetProject(projectName, nil)
    	// ...
    }
    ```

    </details>

3. (optional) Explicitly implement interface methods if it's more convenient, although this might cause embedded method name conflicts

    <details>

    <summary>Example</summary>

    ```go
    func (client *Client) GetBranch(pid any, branch string, options ...gitlab.RequestOptionFunc) (*gitlab.Branch, *gitlab.Response, error) {
    	result, resp, err := client.glClient.Branches.GetBranch(pid, branch, options...)
    	if err != nil {
    		return nil, nil, fmt.Errorf("%w", err)
    	}
    
    	return result, resp, nil
    }
    ```

    </details>

lallevato-lm (approved):

internal/pkg/client/gitlab/fetch.go
```diff
@@ -0,0 +1,247 @@
...
+	domain := url.Hostname
+	if url.Port != "" {
+		domain = fmt.Sprintf("%s:%s", domain, url.Port)
+	}
+
+	baseURL := fmt.Sprintf("https://%s/api/v4", domain)
+	projectName := url.Path
+	branchName := url.Fragment
+
+	gitLabToken := os.Getenv("GITLAB_FETCH_TOKEN")
```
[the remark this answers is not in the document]
jhoward-lm: Before documenting anything, I think it should be just `GITLAB_TOKEN`.

- it's more canonical and consistent with other CLI tools, GitLab's documentation, etc.
- it's reusable for other operations like push

I could also see an argument for prefixing it as `BOMCTL_GITLAB_TOKEN` in case we want to turn it into a command line option or something at some point.

internal/pkg/client/gitlab/fetch.go
```diff
@@ -0,0 +1,247 @@
...
+
+type (
+	ProjectProvider interface {
+		GetProject(
+			any,
+			*gitlab.GetProjectOptions,
+			...gitlab.RequestOptionFunc,
+		) (*gitlab.Project, *gitlab.Response, error)
+	}
+	BranchProvider interface {
```
jhoward-lm, line 43:
```suggestion
	}
	
	BranchProvider interface {
```

internal/pkg/client/gitlab/fetch.go
```diff
@@ -0,0 +1,247 @@
...
+		GetBranch(
+			any,
+			string,
+			...gitlab.RequestOptionFunc,
+		) (*gitlab.Branch, *gitlab.Response, error)
+	}
+	CommitProvider interface {
```
jhoward-lm, line 50:
```suggestion
	}

	CommitProvider interface {
```

internal/pkg/client/gitlab/fetch.go
```diff
@@ -0,0 +1,247 @@
...
+		GetCommit(
+			any,
+			string,
+			*gitlab.GetCommitOptions,
+			...gitlab.RequestOptionFunc,
+		) (*gitlab.Commit, *gitlab.Response, error)
+	}
+	DependencyListExporter interface {
```
jhoward-lm, line 58:
```suggestion
	}

	DependencyListExporter interface {
```

internal/pkg/client/gitlab/fetch.go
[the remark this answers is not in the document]
lallevato-lm: Agreed. I personally lean towards `BOMCTL_GITLAB_TOKEN`, since there is a possibility of a separate GitLab token with different permissions existing on the system which might occupy the variable `GITLAB_TOKEN`.

internal/pkg/client/gitlab/client.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/bomctl/bomctl/internal/pkg/netutil"
+)
+
+type Client struct {
+	Client ClientWrapper
+	Export *gitlab.DependencyListExport
+
+	InitFetch func(*Client) error
+}
```
jhoward-lm, line 36: There's still an extra layer of embedding here. Instead of embedding the interfaces into `ClientWrapper` and then embedding `ClientWrapper` into `Client`, why not just embed them directly into the client?

Instead of a concrete `Export *gitlab.DependencyListExport`, would it not be easier to mock by declaring it as an anonymous embedded `DependencyListExporter`? It should have the same methods, but it adds the ability to reimplement them during tests.

```suggestion
type Client struct {
	ProjectProvider
	BranchProvider
	CommitProvider
	DependencyListExporter

	InitFetch func(*Client) error
}
```

lmphil: Good suggestion, it looks cleaner that way.

As for the `Export` field, it is just a data struct, and interfaces only specify functionality.

internal/pkg/client/gitlab/fetch.go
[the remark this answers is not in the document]
lmphil: I agree. I changed it to the recommended name: `BOMCTL_GITLAB_TOKEN`.

Where can I document that variable? I thought about the `bomctl fetch` help, but I know that we're trying to keep the command client-agnostic.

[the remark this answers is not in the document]
ashearin: As for location, I'd either put it in the fetch section in the README.md, or start an auth section instead. 

Would it make sense to make an issue to match this functionality in the other clients?

ashearin:
Heads up: the go-gitlab package was deprecated....today: https://pkg.go.dev/github.com/xanzy/go-gitlab#section-readme

ashearin:
What am I doing wrong: 
```bash
$ go run . fetch https://www.gitlab.com/hoppr/hoppr@dev
INFO  fetch: Fetching from GitLab URL url=https://www.gitlab.com/hoppr/hoppr@dev
FATAL fetch: failed to save document: adding document: parsing SBOM data: detecting SBOM format: detecting format: unknown SBOM format
exit status 1
```

lmphil:
> What am I doing wrong: (running on your branch obvs)
> 
> ```shell
> $ go run . fetch https://www.gitlab.com/hoppr/hoppr@dev
> INFO  fetch: Fetching from GitLab URL url=https://www.gitlab.com/hoppr/hoppr@dev
> FATAL fetch: failed to save document: adding document: parsing SBOM data: detecting SBOM format: detecting format: unknown SBOM format
> exit status 1
> ```

It was a bug on my part. Sorry. It's fixed now.

ashearin (approved):

README.md
```diff
@@ -154,6 +154,12 @@ Alternatively, use the following syntax to directly fetch the most recent depend
 bomctl fetch https://github.com/bomctl/bomctl
 ```
 
+Or you can use the following syntax to fetch an SBOM from a GitLab repository through the [DependencyListExport web API](https://docs.gitlab.com/ee/api/dependency_list_export.html). Authorization for this command is configured by assigning the value of your GitLab token to the `BOMCTL_GITLAB_TOKEN` environment variable.
```
jhoward-lm, line 157: Avoid "you" statements in documentation

```suggestion
An SBOM may also be fetched from a GitLab repository through the [DependencyListExport web API](https://docs.gitlab.com/ee/api/dependency_list_export.html) by using the following URL format. Authorization for this command is configured by assigning the value of your GitLab token to the `BOMCTL_GITLAB_TOKEN` environment variable.
```

internal/pkg/client/gitlab/client.go
```diff
@@ -0,0 +1,105 @@
...
+	return &netutil.URL{
+		Scheme:   results["scheme"],
+		Hostname: results["hostname"],
+		Port:     results["port"],
+		Path:     results["path"],
+		Fragment: results["branch"],
+	}
+}
+
+func NewGitLabClient(sourceURL, gitLabToken string) *Client {
```
jhoward-lm, line 76: Since `gitlab` and `client` are already part of the import path, how do you feel about renaming this function to just `New`?

Can the token be retrieved from inside this function instead of passing it in? The environment variable should still be configurable within tests.

internal/pkg/client/gitlab/client.go
```diff
@@ -0,0 +1,105 @@
...
+	client := &Client{
+		GitLabToken: gitLabToken,
+	}
```
jhoward-lm, line 79: I would move this to the bottom of the function and create and return it simultaneously. Set a `gitlabToken` variable if you don't want to call `os.GetEnv` more than once

```go
	return &Client{
		GitLabToken:            os.GetEnv("BOMCTL_GITLAB_TOKEN"),
		ProjectProvider:        gitLabClient.Projects,
		BranchProvider:         gitLabClient.Branches,
		CommitProvider:         gitLabClient.Commits,
		DependencyListExporter: gitLabClient.DependencyListExport,
	}
```

You can omit setting `client.Export` to nil because it's the default zero value

internal/pkg/client/gitlab/client.go
```diff
@@ -0,0 +1,105 @@
...
+	if url == nil {
+		return nil
+	}
+
+	domain := url.Hostname
+	if url.Port != "" {
+		domain = fmt.Sprintf("%s:%s", domain, url.Port)
+	}
+
+	baseURL := fmt.Sprintf("https://%s/api/v4", domain)
```
jhoward-lm, line 91: This way is fine, but you could also use the builtin `net/url` to parse it and get this same result directly. Assuming imported as `neturl "net/url"`:

```suggestion
	url, err := neturl.Parse(sourceURL)
	if err != nil {
		return nil
	}

	baseURL := fmt.Sprintf("https://%s/api/v4", url.Host)
```

internal/pkg/client/gitlab/client.go
jhoward-lm: Actually...is this function really needed? I think all of its logic could be moved into `Parse`; the client receiver can be updated in-place since it's a pointer.

From the `New` function in `internal/pkg/client/client.go`:

```go
	for _, client := range []Client{
		&github.Client{},
		&gitlab.Client{},
		&git.Client{},
		&http.Client{},
		&oci.Client{},
	} {
		if url := client.Parse(sbomURL); url != nil {
			return client, nil
		}
```

When `client.Parse` gets called on the gitlab client, it can set everything on the client directly. Just make sure the URL string passes all the parsing checks first, then setup the client fields, etc.

lmphil: I would have done that, but since ideally a function should do only one thing (and this one already parses and returns a URL), I opted for a factory function. What are your thoughts on that? Should I merge them into parse to also initialize the client?

jhoward-lm: What if you just rename `NewGitlabClient` to something like `func(client *Client) initClient` (call it whatever makes sense but make it private) and call it from `Parse`?

lmphil: Sorry. I said "this one" referring to `Parse` when you were talking about `NewGitLabClient` initially. 

I'm OK with renaming `NewGitLabClient` to `new` and calling it from `Parse`. What seems unusual for me is having it initialize the instance **_and_** return a URL struct, because, from my experience outside Go, when a parse method is used to initialize an object it means "parse this data into the object", but here that would mean "parse this string into a URL struct **_and_** also initialize my "Client" object.

lmphil: I went ahead and implemented the change, but I'd like to discuss the issue I outlined above. I think we can make the way we implement clients in Bomctl a bit cleaner.

ashearin (approved):

internal/pkg/client/gitlab/client.go
```diff
@@ -0,0 +1,104 @@
...
+type Client struct {
+	ProjectProvider
+	BranchProvider
+	CommitProvider
+	DependencyListExporter
+	Export      *gitlab.DependencyListExport
+	GitLabToken string
+}
+
+func (client *Client) init(sourceURL string) error {
```
jhoward-lm, line 42: Is this still going to go into `PrepareFetch`?

internal/pkg/client/gitlab/client.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := client.init(rawURL); err != nil {
+		return nil
+	}
+
+	return &netutil.URL{
+		Scheme:   results["scheme"],
+		Hostname: results["hostname"],
+		Port:     results["port"],
+		Path:     results["path"],
+		Fragment: results["branch"],
```
ashearin, line 102: Doesn't this diverge from the other parsing functions? git client uses fragment for the path within the repo, and the branch/ref was added to the path with a @ prefix, like `<hostname>/repo/path@branch#path/to/file.json`

Only concerned about UI consistency

README.md
```diff
@@ -154,6 +154,12 @@ Alternatively, use the following syntax to directly fetch the most recent depend
 bomctl fetch https://github.com/bomctl/bomctl
 ```
 
+An SBOM may also be fetched from a GitLab repository through the [DependencyListExport web API](https://docs.gitlab.com/ee/api/dependency_list_export.html) by using the following URL format. Authorization for this command is configured by assigning the value of your GitLab token to the `BOMCTL_GITLAB_TOKEN` environment variable.
+
+```shell
+bomctl fetch https://www.gitlab.com/PROJECT/REPOSITORY@BRANCH
```
ashearin, line 160: Does this work? based on the parse function and my other comment, shouldn't this be:

`bomctl fetch https://www.gitlab.com/PROJECT/REPOSITORY#BRANCH`

internal/pkg/client/gitlab/client.go
ashearin: I understand it's much easier to do it this way when parsing. This is something I'm looking at in my URL parsing proof of concepts

internal/pkg/client/gitlab/fetch.go
```diff
@@ -0,0 +1,198 @@
...
+	branch, response, err := client.GetBranch(project.ID, branchName)
+	if err != nil {
+		return fmt.Errorf("failed to get branch: %w", err)
+	}
+
+	if err := validateHTTPStatusCode(response.StatusCode); err != nil {
+		return err
+	}
+
+	commit, response, err := client.GetCommit(project.ID, branch.Commit.ID, nil)
```
ashearin, line 107: Should we be able to handle a user providing a commit versus always a branch?

Might be a future improvement, but was just thinking about it

internal/pkg/client/gitlab/client.go
jhoward-lm: I think for consistency we should use `@` with the value stored in the `GitRef` field, and the capture group should also be named `gitRef`

internal/pkg/client/gitlab/client.go
lmphil: Yes

README.md
lmphil: We're using the fragment when we parse the provided URL string into a `netutil.URL` with the `Parse` method. The user is still expected to use the `@` delimiter, but internally, we're using the fragment because there's no other field in `netutil.URL`.

internal/pkg/client/gitlab/fetch.go
lmphil: Hmm. Could be a good idea, although right now I cannot think of a use case where a user might want the SBOM for a commit that is not the latest one in a branch. Good use case for discussion.

jhoward-lm: Also need to make sure the API supports that. IIRC it only supports `ref` and `tag` in the request URL params. `ref` might accept a commit SHA, but I don't know off the top of my head. @lmphil does this PR support specifying a tag?

README.md
jhoward-lm: `GitRef` seems perfectly adequate to me

internal/pkg/client/gitlab/fetch.go
jhoward-lm: I don't think we should expect the user to provide a pipeline ID, the goal of this client is to get that pipeline ID for them automatically based on a git ref

jhoward-lm (approved):

internal/pkg/client/gitlab/fetch.go
ashearin: To be resolved in #246

PR #259, opened 2025-01-09
refactor: Parsing adr poc url

ashearin:
Proof of concept for 2 items:
- unified cmd url parsing
- using net/url parsing to remove custom parsing code and rely on standard format

The URL standard is the closest to what bomctl currently uses and was the logical first choice for a Proof of Concept.  The command URL is largely the same as currently utilized, however any custom elements are being removed and moved into a url compliant field. example below reworks this command url in the currently implemented format:
    - `https://username:password@github.com:12345/bomctl/bomctl.git@main#sbom.cdx.json`
```
URL: [scheme:][//[userinfo@]host][/]path[?query][#fragment]

https://username:password@github.com:12345/bomctl/bomctl.git?ref=main#sbom.cdx.json

scheme: https
userinfo: [username, password]
host: github.com:12345 (parse port with url.Port())
path: bomctl/bomctl.git
fragment: sbom.cdx.json
Query: ref=main (parse query with url.Query() and get value with query.Get("ref"))
```

Main difference is the `@main` used to designate the branch is moved into the query field. Another option is removing it entirely and creating a branch flag and defaulting to the default branch if a branch is not provided (although this may only be applicable to the git client)


internal/pkg/client/client.go
```diff
@@ -47,3 +55,85 @@ type (
...
+		//nolint:wrapcheck
+		return oci.Init(parsedURL)
+	default:
+		return nil, fmt.Errorf("%w", errors.ErrUnsupported)
+	}
+}
+
+func registrySlice() []string {
+	// maybe check for env variable like BOMCTL_REGISTRY or something and append to slice if it exists
+	return []string{"registry", "harbor", "ghcr.io", "gcr.io", "docker.io"}
```
ashearin, line 138:
```suggestion
	return []string{"quay.io", "ghcr.io", "gcr.io", "docker.io"}
```
Remove items that are too general and likely to cause collision/mistaken match, add only known registries

Issue #256, opened 2025-01-08
Add external reference support for oci client


pkwiatkowski1:
Updated external references provided in the annotations field on push.

PR #266, opened 2025-01-13
feat: add gitlab push to generic package registry

lmphil:
## Description

Adds GitLab Push functionality to push an SBOM to a project's generic package repository. 

Fixes issue #256

## Type of change

## How Has This Been Tested?

- [x] Tested locally.
- [x] Tested using unit tests.

## Checklist


internal/pkg/client/gitlab/client.go
```diff
@@ -17,7 +17,7 @@
 // limitations under the License.
 // -----------------------------------------------------------------------------
 
-package gitlab
+package gitlab //nolint:revive
```
jhoward-lm, line 20: We don't want to wholesale disable an entire linter, and this directive applies to every file in this package.

Either make some of these types private/unexported (preferred), or **_as a last resort_**, don't use `//nolint:revive`, but instead a revive-specific directive including the specific rule.

internal/pkg/client/gitlab/client_test.go
```diff
@@ -53,14 +67,31 @@ type (
...
 )
 
 //revive:disable:unchecked-type-assertion
 
+func (mpp *mockGenericPackagePublisher) PublishPackageFile(
+	pid any,
+	packageName, packageVersion, fileName string,
+	content io.Reader,
+	opt *gogitlab.PublishPackageFileOptions,
+	options ...gogitlab.RequestOptionFunc, //nolint:gocritic,revive
```
jhoward-lm, line 83: Don't use `//nolint:revive`, instead use a revive directive to disable a specific linter, e.g. 

```go
//revive:disable:flag-parameter
```

Check to see if `gocritic` has a similar directive you can use. Please address all instances of this comment throughout the file

jhoward-lm (changes requested):

internal/pkg/client/gitlab/client.go
lmphil: Fixed using your `internal_test.go` solution below.

internal/pkg/client/gitlab/client_test.go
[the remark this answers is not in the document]
lmphil: All set.

internal/pkg/client/gitlab/client_test.go
lmphil: All set. There's no similar way to disable specific `gocritic` rules.

internal/pkg/client/gitlab/client.go
```diff
@@ -70,5 +138,14 @@ func (client *Client) Parse(rawURL string) *netutil.URL {
 		Port:     results["port"],
 		Path:     results["path"],
 		GitRef:   results["gitRef"],
+		Query:    results["query"],
 	}
 }
+
+func validateHTTPStatusCode(statusCode int) error {
+	if statusCode < http.StatusOK || http.StatusMultipleChoices <= statusCode {
```
jhoward-lm, line 146: nit: slightly more readable to me (subjective). Did the linter suggest this?

```suggestion
	if statusCode < http.StatusOK || statusCode >= http.StatusMultipleChoices {
```

internal/pkg/client/gitlab/push.go
```diff
@@ -19,16 +19,143 @@
...
+func (client *Client) PreparePush(pushURL string, _opts *options.PushOptions) error {
+	gitLabToken := os.Getenv("BOMCTL_GITLAB_TOKEN")
+
+	url := client.Parse(pushURL)
+
+	host := url.Hostname
+
+	if url.Port != "" {
+		host = fmt.Sprintf("%s:%s", host, url.Port)
+	}
```
jhoward-lm, line 52: I think `url.Host` should handle with or without port

internal/pkg/client/gitlab/internal_test.go
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// -----------------------------------------------------------------------------
+
+package gitlab
+
+type (
+	StringWriter = stringWriter
+)
```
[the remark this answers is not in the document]
lmphil: all set

internal/pkg/client/gitlab/client.go
lmphil: All set. The linter had no suggestions on comparison order.

internal/pkg/client/gitlab/push.go
lmphil: `netutil.URL` doesn't have a `Host` field. Do you mean I should use just `url.Hostname` without the port?

jhoward-lm: Oh, my mistake. I thought it was a `net/url.URL`

jhoward-lm (approved):

internal/pkg/client/gitlab/push.go
```diff
@@ -19,16 +19,143 @@
...
+
+	sbom, err := backend.GetDocumentByIDOrAlias(id)
+	if err != nil {
+		return fmt.Errorf("%w", err)
+	}
+
+	uuidRegex := regexp.MustCompile(`urn:uuid:([\w-]+)`)
+	uuidMatch := uuidRegex.FindStringSubmatch(id)
+	sbomFilename := uuidMatch[1]
```
ashearin, line 85: Does this work with spdx IDs? Also should we be checking before we try to access the indices?

```
$ go run . push -v https://anchore.com/syft/file/bomctl_0.3.0_linux_amd64.tar.gz-1b838d44-9d3c-47d0-9f7f-846397e701fa#DOCUMENT "https://gitlab.com/Ashearin/another-
access-test?package_name=really_cool_package&package_version=v1.2.7"
INFO  push: Pushing document id=https://anchore.com/syft/file/bomctl_0.3.0_linux_amd64.tar.gz-1b838d44-9d3c-47d0-9f7f-846397e701fa#DOCUMENT
INFO  push: Pushing to GitLab URL url="https://gitlab.com/Ashearin/another-access-test?package_name=really_cool_package&package_version=v1.2.7"
INFO  push: Adding file id=https://anchore.com/syft/file/bomctl_0.3.0_linux_amd64.tar.gz-1b838d44-9d3c-47d0-9f7f-846397e701fa#DOCUMENT
exit status 1
```

internal/pkg/client/gitlab/client.go
```diff
@@ -58,7 +119,14 @@ func (client *Client) Parse(rawURL string) *netutil.URL {
 	}
 
 	// Ensure required map fields are present.
-	for _, required := range []string{"scheme", "hostname", "path", "gitRef"} {
+	requiredFields := []string{
+		"scheme",
+		"hostname",
+		"path",
+		// "gitRef", // Required if Fetch only
+		// "query",  // Required if Push only
```
ashearin, line 127: Do we want to deviate how we're expecting users to pass in the url between this and the other clients? (really just git)

`https://gitlab.com/<PROJECT_NAME>?package_name=<PACKAGE_NAME>&package_version=<PACKAGE_VERSION>`
vs
`"https://gitlab.com/<PROJECT_NAME>@<PACKAGE_VERSION>#<PACKAGE_NAME>"`

I realize we may be switching eventually, but we should stay consistent until we decide on a path forward

internal/pkg/client/gitlab/client.go
```diff
@@ -44,8 +105,8 @@ func (*Client) Name() string {
 func (*Client) RegExp() *regexp.Regexp {
 	return regexp.MustCompile(fmt.Sprintf("(?i)^%s%s%s$",
 		`(?P<scheme>https?|git|ssh):\/\/`,
-		`(?P<hostname>[^@\/?#:]+gitlab[^@\/?#:]+)(?::(?P<port>\d+))?/`,
-		`(?P<path>[^@#]+)@(?P<gitRef>\S+)`))
+		`(?P<hostname>[^@\/?#:]*gitlab[^@\/?#:]+)(?::(?P<port>\d+))?/`,
+		`(?P<path>[^@?#]+)(?:@(?P<gitRef>[^?#]+))?(?:\?(?P<query>[^#]+))?`))
```
ashearin, line 109: I was surprised that the commands dont seem to work if the URL is not in quotes:

```
$ go run . push urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db https://gitlab.com/Ashearin/another-access-test?package_name=different_package&package_version=0.0.0
[1] 45754
usmc02g900fmd6r:philippe-bomctl e386920$ INFO  push: Pushing document id=urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db
INFO  push: Pushing to GitLab URL url="https://gitlab.com/Ashearin/another-access-test?package_name=different_package"
INFO  push: Adding file id=urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db
FATAL push: failed to push sbom: PUT https://gitlab.com/api/v4/projects/52635138/packages/generic/different_package/f360ad8b-dc41-4256-afed-337a04dff5db.json: 400 {error: file is missing}
exit status 1
\\ **(had to manually stop the process)
[1]+  Exit 1                  go run . push urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db https://gitlab.com/Ashearin/another-access-test?package_name=different_package
$ 
```
but with quotes: 
```
$ go run . push urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db "https://gitlab.com/Ashearin/another-access-test?package_name=different_package&package_version=0.0.0"
 
INFO  push: Pushing document id=urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db
INFO  push: Pushing to GitLab URL url="https://gitlab.com/Ashearin/another-access-test?package_name=different_package&package_version=0.0.0"
INFO  push: Adding file id=urn:uuid:f360ad8b-dc41-4256-afed-337a04dff5db
```
Is this intentional?

ashearin (changes requested):
Some things I noticed playing around with it. 

Also if we keep the currently implemented expected form for this clients, we need to update docs to make it easily usable.

internal/pkg/client/gitlab/client.go
lmphil: The symbols `?` and `&` have special meaning in the CLI. That might be the reason why it only works with quotes.

internal/pkg/client/gitlab/client.go
lmphil: We can do that, to keep it consistent, although, the semantics will be incorrect. We're storing what's after `@` as `gitRef`, and the package name and package version are the identifiers for the package we will be pushing/creating.

What do you suggest @jhoward-lm?

internal/pkg/client/gitlab/push.go
lmphil: Good catch. I have to test with SPDX IDs. I made the assumption that they all had the same format because they were in the same field.

internal/pkg/client/gitlab/client.go
ashearin: Ah, that makes sense. I couldn't think of why it would make a difference. 

Also good to know for the Parsing ADR

internal/pkg/client/gitlab/client.go
ashearin: I think keeping it consistent would also eliminate the need for special variables for push and fetch in the gitlab client. Just use gitref and fragment.

internal/pkg/client/gitlab/client.go
jhoward-lm: I think just `&` is significant. In your output `[1] 45754` means the shell spawned a background job

ashearin: I guess the only thing here would be documenting this, also kinda makes me think we shouldn't be using either of those in cmd urls (or at least expecting them). Might be a reason to keep with the current paradigm.

internal/pkg/client/gitlab/push.go
lmphil: SPDX IDs are now working

internal/pkg/client/gitlab/client.go
lmphil: Updated to use the suggested URL convention.

internal/pkg/client/gitlab/client.go
lmphil: Updated the README to document GitLab push usage.

ashearin (approved):

